PR #36, opened 2015-01-06
Fix vpn03 firewall rules

andrenarchy:
The vpn03 firewall rules were non-functional since the switch from ips to domains in the vpn03 config. This PR fixes the firewall rules and removes some unnecessary code in the freifunk-berlin-openvpn-defaults package. Note that the vpn03-related firewall rules are now set in the uci-defaults file of the freifunk-berlin-openvpn-defaults package.


andrenarchy:
By the way: this PR should also fix the OLSR error messages mentioned in freifunk-berlin/firmware#171 and freifunk-berlin/firmware#192.

booo:
With my changes I think this is good to go. I will test again in a few minutes and merge if tests are good.

booo:
Tests are positive on a WDR4900. Merged.

PR #37, opened 2015-01-18
[olsrd-defaults] set SmartGatewayThreshold=50

andrenarchy:
Without SmartGatewayThreshold being set, the smart gateway endpoint is only
changed when it disappears. With SmartGatewayThreshold=50, it picks
another gateway if its ETX is 50% or less of the current gateway's ETX.

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -52,6 +52,15 @@ update_luci_statistics_config() {
   /etc/init.d/luci_statistics enable
 }
 
+update_olsr_smart_gateway_threshold() {
+  # set SmartGatewayThreshold if not set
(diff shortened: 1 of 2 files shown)


andrenarchy:
I forgot: this has been tested on several devices. :)

booo:
Can you describe your testing procedure so we can replicate the tests?

booo:
We should add this to the migration script!

andrenarchy:
Test procedure:

1. make sure you have 2 SmartGateways A and B with ETX(A) > 2*ETX(B) in your OLSR mesh (i.e., B is the best gateway based on ETX alone).
2. connect test client with set SmartGatewayThreshold to OLSR mesh.
3. let SmartGateway B disappear -> test node should switch to A.
4. reactivate SmartGateway B -> test node should switch back to B (won't happen without SmartGatewayThreshold).

andrenarchy:
+1 for the migration script! :)

booo:
@andrenarchy are you going to work on this?

cholin:
Merged manually.

PR #39, opened 2015-01-31
Exclude encrypted/wan interfaces in OWM info

sarumpaet:
See https://github.com/freifunk-berlin/firmware/issues/206 .

This patch is intended to prevent owm.lua from pushing information about a) private encrypted wifi networks and b) the local WAN port configuration to the public.

Ultimatively, I think there should be a user-configurable whitelist instead of a blacklist, and some "show me what information gets published beforehand" functionality, but this is way too much effort with libremap coming.

### DIFF
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -352,6 +352,10 @@ function get()
 			return
 		end
 		local name = vif['.name']
+		if ('wan' == name) or ('wan6' == name) then
+			-- fingers off wan as this will be the private internet uplink
+			return
+		end
 		local net = netm:get_network(name)
 		local device = net and net:get_interface()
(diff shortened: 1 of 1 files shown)


utils/luci-app-owm/luasrc/owm.lua
```diff
@@ -344,6 +351,10 @@ function get()
 		if 'lo' == vif.ifname then
 			return
 		end
+		if 'br-wan' == vif.ifname then
```
andrenarchy: I think `vif.name` should be checked against `wan` and `wan6` here.

andrenarchy:
Thanks for the PR! The encryption check looks good. Can you update the PR with an adapted wan check?

sarumpaet:
Thanks. There. :)

andrenarchy:
Works nicely on my router.

andrenarchy:
Thx for your contribution!

PR #40, opened 2015-02-10
[firewall-defaults] remove iptables dependencies

cholin:
fw3 (firewall of OpenWRT) is capable of setting raw network devices since
v2.58. Remove iptables dependency and add SmartGateway tunnels to zone_freifunk
via uci.


cholin:
Unfinished. I'll will make another in the next days.

PR #41, opened 2015-02-11
[ffwizard-berlin] use move instead of copy after upload of key/cert

booo:
Before this change we copied the key/cert files from the upload directory to the
/etc/openvpn directory. This results in a waste of disk space. If we use move
instead we do not keep redundant information on disk and save some space.

Thanks to Patrick Grimm for pointing this out.


booo:
Works for me. Tested on a WDR4900.

PR #47, opened 2015-02-23
Fix collectd memory leak hotfix

sarumpaet:
This is supposed to remove the noop line put into crontab by the 0.1.0 migration and insert a working command. Also, reloading every 15 minutes as at least some WR1043 seem not to last even 30 minutes. Do not merge yet, will do more testing.

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -52,13 +52,15 @@ update_luci_statistics_config() {
   /etc/init.d/luci_statistics enable
 }
 
-update_crontab_memory_leak_fix() {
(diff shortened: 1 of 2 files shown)


utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
```diff
@@ -79,10 +81,11 @@ migrate () {
     update_dhcp_lease_config
     update_wireless_ht20_config
     update_luci_statistics_config
-    update_crontab_memory_leak_fix
     update_olsr_smart_gateway_threshold
   fi
 
+  update_collectd_memory_leak_hotfix
+
```
cholin: Why don't you create a new `if semverLT ${OLD_VERSION} "0.2.0"; then`-block and put `update_collectd_memory_leak_hotfix` in there?

cholin:
Rest looks fine to me. If you've time you could integrate my suggestion otherwise just merge this. :smile:

PR #48, opened 2015-03-11
[olsrd-defaults] set txtinfo port to 2006 (default)

sarumpaet:
Both olsrd-txtinfo and collectd-olsrd default to port 2006; this changes our config to that default. http://www.olsr.org/?q=txtinfo_plugin https://collectd.org/wiki/index.php/Plugin:olsrd

This fixes https://github.com/freifunk-berlin/firmware/issues/218 . Do we need migration?


booo:
I think we should add line to the migration script for this. @sarumpaet can you do this?

PR #49, opened 2015-04-20
Changes to support routers with 4MB flash

sarumpaet:
1. The Assistent doesn't really depend on luci-mod-freifunk apart from patching into its menu entry. This removes the dependency: On small routers, luci-mod-freifunk can then be omitted.

2. sysupgrade copies all files in /etc/config to the overlay during the upgrade. Change migration to delete any duplicated files; this saves a dozen or so kb typically. Note: Patching sysupgrade itself would be better from a space-preserving perspective (as currently, the overlay has to hold the duplicate files briefly), but it would change the upgrate semantics from "keep config as-is" to "update any non-modified config files to ROM default".

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -98,6 +98,9 @@ migrate () {
   uci commit
 
   log "Migration done."
+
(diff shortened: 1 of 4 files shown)


sarumpaet:
https://github.com/freifunk-berlin/firmware-packages/commit/dfd61e13c0d2ed7f7c69552d6595d1ddf2b1379c fixes a small bug introduced by ab5216a (which caused the Assistent page to show up instead of the luci-mod-freifunk status page as first page in the admin view).

PR #50, opened 2015-05-04
Assistent/FF-Wizard updates

sarumpaet:
See the commit descriptions. The main goal is to make the Assistent more of a replacement of luci-mod-freifunk: With these changes, the Assistent can get re-run rather quickly in order to update settings. No re-upload of VPN certs necessary. Contact information is updated immediately when submitting the first page.

Removing the ISP bandwidth defaults removes a lot of code; the defaults were unmaintained and badly named (who'd have guessed that "VDSL 50" means 50MBit/s up AND down? Does this even exist?). Also see https://github.com/freifunk-berlin/firmware/issues/173 and https://github.com/freifunk-berlin/firmware/issues/213 .

Not specifying bandwidth sets "customBW" to 0 now and is supposed to disable QoS.

I did a bit of testing, but more testing especially concerning VPN/bandwidth changes would be good.


utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/ffwizard.lua
```diff
@@ -36,28 +36,26 @@ end
 
 function configureQOS()
   if sharenet == "1" then
-    -- values have to be in kilobits/seconed
-    local up = 128
-    local down = 1024
-    if uci:get("ffwizard", "settings", "customBW") == "0" then
-      local usersBandwidthUp = bandwidths[uci:get("ffwizard", "settings", "usersBandwidth")].up
-      local usersBandwidthDown = bandwidths[uci:get("ffwizard", "settings", "usersBandwidth")].down
-      local shareBandwidth = uci:get("ffizward", "settings", "usersBandwidth") or 100
```
sarumpaet: "ffizward" ... or 100. Sneaky! ;)

sarumpaet:
I made bandwidth settings mandatory again. While we can disable the traffic limiter, SmartGateway needs to know the available bandwidth (it defaults to 1024/128 Kbit/s which is a bit 90s).

sarumpaet:
On an 841, this (build 589) works fine.
Once someone merges this, https://github.com/freifunk-berlin/firmware/pull/252 and https://github.com/freifunk-berlin/firmware-packages/pull/48 , I'd say 0.1.1 is ready. @booo, @andrenarchy? I think we should release 0.1.1 before the WCW...

PR #51, opened 2015-05-15
[freifunk-berlin-openvpn-files] keep content of /etc/openvpn on sysup…

booo:
…grade

This should fix https://github.com/freifunk-berlin/firmware/issues/238

We move the ffvpn-up.sh script to the /lib/freifunk directory. This way we
exclude the script from the backup of /etc/openvpn. Other communities already
use /lib/freifunk as directory for code so we use it instead of
/usr/lib/freifunk.

This is untested...

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -85,6 +85,10 @@ add_openvpn_mssfix() {
   uci set openvpn.ffvpn.mssfix=1300
 }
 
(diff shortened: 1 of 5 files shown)


sarumpaet:
I think migration is needed for this; uci-defaults are run once after reflashing in general, but the guard statement will exit since the script has been run for the previous firmware version already.

booo:
You are totally right. I will add a line to the migration script.

booo:
@sarumpaet Can you take a look at this and merge?. I updated the pull request. I have tested this and it works for me on 0.1.2 -> 0.2.0-apha (wdr3500).

PR #54, opened 2015-09-12
[uhttpd-defaults]: do not force a redirect to https

booo:
The assumption is that the forced redirect is irritating for the average users
because the self signed certificate is not trusted by the browser. The user has
to add the certificate by hand. We assume that is to much for the average user.


sarumpaet:
Thanks! I think we need migration for this, too.

booo:
As long as the default configuration for uttpd does not contain a forced redirect I don't see a reason why. Old configurations do not contain the redirect option so we don't have to change them.

PR #57, opened 2015-09-26
[ff-berlin-statistics-defaults]: change ping host for collectd

booo:
We change the dns entry for the ping plugin to a more flexible dns entry. At the
moment the entry only points to a ipv4 address.

Addresses: https://github.com/freifunk-berlin/firmware/issues/253

TODO:

* [x] create a migration step

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -105,6 +105,11 @@ add_firewall_rule_vpn03c() {
   fi
 }
 
+update_collectd_ping() {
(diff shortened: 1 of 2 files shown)


utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
```diff
@@ -105,6 +105,11 @@ add_firewall_rule_vpn03c() {
   fi
 }
 
+update_collectd_ping() {
+ uci set luci_statistics.collectd_ping.Interval=30
```
booo: Why do we use 30 here and in the defaults 10?

sarumpaet:
I fixed the 30 seconds typo.

PR #58, opened 2015-10-08
[ffwizard] [migration] use ffvpn for QoS rather than wan

sarumpaet:
this fixes https://github.com/freifunk-berlin/firmware/issues/257


booo:
Looks good to me. Tested. Works. Merged.

PR #59, opened 2015-10-18
[luci-app-ffwizard-berlin] Remove existing certificate files

buschjost:
Allows to use the wizard to change certificates. Before this failed due to already existing files in the destination of the following move.

This fixes the issue https://github.com/freifunk-berlin/firmware/issues/287 and is based on the patch provided there.

### DIFF
--- a/utils/luci-app-ffwizard-berlin/htdocs/freifunk/assistent/assistent.css
+++ b/utils/luci-app-ffwizard-berlin/htdocs/freifunk/assistent/assistent.css
@@ -17,3 +17,7 @@
 #map {
   margin-top:20px;
 }
+
+.cbi-input-image {
+	width: 55px
+}
--- a/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/shareInternet.lua
+++ b/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/shareInternet.lua
@@ -1,7 +1,6 @@
 local uci = require "luci.model.uci".cursor()
 local fs = require "nixio.fs"
 local tools = require "luci.tools.freifunk.assistent.tools"
-local html = require "luci.http"
 
 f = SimpleForm("ffvpn","","")
 f.submit = "Next"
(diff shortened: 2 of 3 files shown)


sarumpaet:
I think that section gets run even if the user didn't choose to re-upload certs, so possibly that change would delete existing certs and leave OpenVPN defunct in that case. Please check.

buschjost:
I have fixed the issue by only removing and moving the key and certificate, when both certificate and key are present from a new upload.

booo:
I will test and merge in the next days. Looks good so far.

booo:
If there is a key and cert in /etc/openvpn I can't upload another version. @buschjost did you test this?

sarumpaet:
@booo are you sure you tested the right version? The code changes look correct and aren't very complicated. I'd just merge.

booo:
@sarumpaet no but someone else should definitely test this too.

SvenRoederer:
Checked this too and it did not work.
After clicking on button "VPN-Schlüssel erneute hochladen" the upload-boxes come up and I can select the new files. But the "Back" and "Next" buttons are not functioning.

buschjost:
@boo I have found the issue you are experiencing. It is a problem that exists due to how the upload form is constructed with the 'Reupload button' behavior. So the new files do never reach the my changes in the normal case (when the .crt is found in the filesystem - when only the key exists everything works as expected because you see the initial form behaviour).
I tried different variants, but it looks like that I can not keep the 'Reupload Button behavior' as it currently is and make it work. (Problems are: I would have to submit a hidden dummy value in the form which seems to be not possible. Alternatively I would have to alter the form action to append a query parameter which I can not because it is set to REQUEST_URI through upstream packages).  

**My solution** would be to remove this Re-Upload Button. I would always directly show the upload forms for the files with one important difference: As soon as files already exists in the filesystem the upload form for this file would become *optional* instead of *required*. I coud also add a text indicating these as optional in this case.

@SvenRoederer I am not able to reproduce your issue. 

Would this changed behavior of the form be ok for you and get merged? Then I update my PR.

buschjost:
I updated the branch with a new uploaded behavior. I am not really happy with it, but it seems to me the best I can get with my knowledge of lucy and the constraints of the lucy system. I removed the big reupload files button and now utilize the standard lucy behavior with an 'replace' button that effectively deletes the original file from the filesystem. Take a look at this screen: 

bildschirmfoto 2016-01-12 um 00 06 30

Feel free to just kill this pull request.

booo:
@SvenRoederer did you check on the current solution this?

SvenRoederer:
nope, needs to be verified

booo:
If I push the replace button I get redirected to the next page because the form is valid even tough I have not uploaded anything.

SvenRoederer:
maybe leave this for next release, as chainging the VPN-key is a uncommon thing and will not affect many users

bobster-galore:
I think the average user will only have this issue if a common vpn is closed. In that case I suggest that we give instructions on the mailing list / wiki. 
People who tinker with several routers will do it probably on the command line.
To me it seems we can do that in a later release (after 0.2.0).

sarumpaet:
@bobster-galore @SvenRoederer We got quite a few reports of people having trouble with this. E.g., people got a second router, test-configured this with the same cert as the first router (because getting a new cert takes a while), then tried to replace the cert with the new one, and ended up with two routers that cannot be used concurrently, without any indication that anything went wrong. Also, I've seen routers with broken/swapped cert/key files. Fixing that should be possible from the wizard really.

It should be possible to add a "reupload cert" button that just deletes the current cert/key, then redirects to the `shareInternet` page (without the `reupload` parameter) again.

bobster-galore:
@andrenarchy @sarumpaet @SvenRoederer I see that there is a need, especially when a button does not what's written on it. I suggest that this issue will be solved in the new wizard. Could this be a way to solve it?

SvenRoederer:
As this issue seems to be in the assistent since some time (at least kathleen 0.1.2) and the new wizard is making progress, I suggest to keep this problem.
There seems to be no "ready for use code" and I prefer to push this 0.2.0-release as long as there is nothing more broken than in the last release. So we can end up in "at least more recent hardware support and stable code (olsr, collectd)". The open "old" issues we can fix in the next release

sarumpaet:
@SvenRoederer Really, if we get bug reports of this kind ("xyz just doesn't work") we either fix it or disable the feature completely. We don't keep bugs "because they have been there for some time".

@bobster-galore I doubt the new wizard will be ready soon, but who knows. It's been dormant for many months until the recent hack event.

Anyways, I just committed https://github.com/freifunk-berlin/firmware-packages/commit/20f1ce26a21e9063fd363da55bcdf3ec40b0b072 which should fix the issue. Four lines of code vs. dozens of comments, we're SUCH a good programming team. ;-)

PR #61, opened 2015-11-16
remove "latlon" from uci: system.system

SvenRoederer:
- this removes the uci-setting "system.system.latlon"
- "latlon" seems to be only historic, also by checking the whole src-tree (relevant for Freifunk) there seems to be no active use of the parameter; only found some fallback in case "system.system.latitude" and "system.system.longitude" are not defined
- code seems to work fine on my systems
- this closes #307

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -110,6 +110,13 @@ update_collectd_ping() {
  uci set luci_statistics.collectd_ping.Hosts=ping.berlin.freifunk.net
 }
 
+fix_qos_interface() {
+  for rule in `uci show qos|grep qos.wan`; do
+    uci set ${rule/wan/ffvpn}
+  done
+  uci delete qos.wan
+}
+
 migrate () {
   log "Migrating from ${OLD_VERSION} to ${VERSION}."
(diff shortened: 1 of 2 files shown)


booo:
The olsrd init script uses the latlon option:

https://github.com/openwrt-routing/packages/blob/f4b5a5c6084ca327c54a339938854cc84100580a/olsrd/files/olsrd.init

Can you take a closer look at this again? I do think it makes sense to just have one option. I'm in favor for latitude and longitude options.

SvenRoederer:
my understanding is that in Line57 is tried to lookup latlon and if that failsin teh next section SYSTEM_LAT & SYSTEM_LON are taken from uci-setting "lat" &  "lon".
so the first lookup will return no result and the 2nd wil return the correct values. I think it's ok to have the check for the old setting in the code for "legacy"- resaons.

SvenRoederer:
I have seen the code around https://github.com/openwrt-routing/packages/blob/f4b5a5c6084ca327c54a339938854cc84100580a/olsrd/files/olsrd.init#L57 
It tries at first to get the LAT & LON from the combined field "LATLON" and if the fails (line 64) it takes the values from LAT and LON directly.

booo:
@SvenRoederer Why did you close this?

SvenRoederer:
indeed, it looks like I did ... but maybe it's related to a cleanup of my fork ....

PR #63, opened 2015-12-08
Correcting owm deps

SvenRoederer:
- luci-app-owm-cmd provides the "owm.lua" to update the map. but the cronjob for this was setup by package "luci-app-owm-gui".
- also include the setup for cron (/etc/uci-defaults/owm) directly in postinst-hook of package

### DIFF
--- a/utils/luci-app-owm/files/owm-defaults
+++ b/utils/luci-app-owm/files/owm-defaults
@@ -1,7 +1,0 @@
-#!/bin/sh
-test -f /etc/crontabs/root || touch /etc/crontabs/root
-SEED="$( dd if=/dev/urandom bs=2 count=1 2>&- | hexdump | if read line; then echo 0x${line#* }; fi )"
-MIN="$(( $SEED % 59 ))"
-grep -q "owm.lua" /etc/crontabs/root || echo "$MIN * * * *	/usr/sbin/owm.lua" >> /etc/crontabs/root
-/etc/init.d/cron restart
-
--- a/utils/luci-app-owm/files/owm-defaults
+++ b/utils/luci-app-owm/files/owm-defaults
@@ -1,0 +1,7 @@
+#!/bin/sh
+test -f /etc/crontabs/root || touch /etc/crontabs/root
+SEED="$( dd if=/dev/urandom bs=2 count=1 2>&- | hexdump | if read line; then echo 0x${line#* }; fi )"
+MIN="$(( $SEED % 59 ))"
+grep -q "owm.lua" /etc/crontabs/root || echo "$MIN * * * *	/usr/sbin/owm.lua" >> /etc/crontabs/root
+/etc/init.d/cron restart
+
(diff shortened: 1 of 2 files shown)


booo:
Did you test this?

SvenRoederer:
package installed w/o problems and setup the cron-job. my node then popped up in the map again.
Tested with the version having the separate "etc/uci-default/owm" file.But checked the result in the postinst-skript.

Will test on my router with final version ...

booo:
After proper testing I suggest we merge this. @andrenarchy any comments on this?

SvenRoederer:
package installs without problems; setup of crontab is ok, pulls dependency to "luci-app-owm"

sarumpaet:
uci-defaults files get applied at boot time (also after flashing), then deleted. postinst gets applied after package installation. You should test not (only) by installing the package but also with a clean new image. Chances are these changes break firmware images that have the package pre-installed (not sure about this; possibly the modified crontab ends up in SquashFS then which might be fine).

The line
> ( . /etc/uci-defaults/owm ) && rm -f /etc/uci-defaults/owm
probably was just an additional thing to make sure the cronjob gets set up at pkg installation time (without reboot needed).

@SvenRoederer why did you touch this in the first place? Is something broken?

SvenRoederer:
@sarumpaet found that "uci-default/owm" is part of packge "luci-app-owm-cmd" but it gets called in postinst of package "luci-app-owm-gui". there was no dependency between these packages, so "-gui" gave errors during installation.

SvenRoederer:
reverted the inclusion of "uci-defaults/owm" directly into Makefile, as this will break the correct setup of crontab when package is bundled into firmware

stargieg:
hi
use crontab on the right way IMHO
```
crontab -l | grep -q 'owm.lua' || crontab -l | { cat; echo "$MIN * * * *	test -e /usr/sbin/owm.lua && /usr/sbin/owm.lua"; } | crontab -
```

SvenRoederer:
I was also thinking of this, but doing it this way will use more memory in flash. So I decided for the smaller footprint.

utils/luci-app-owm/files/owm-defaults
```diff
@@ -2,6 +2,6 @@
 test -f /etc/crontabs/root || touch /etc/crontabs/root
 SEED="$( dd if=/dev/urandom bs=2 count=1 2>&- | hexdump | if read line; then echo 0x${line#* }; fi )"
 MIN="$(( $SEED % 59 ))"
-grep -q "owm.lua" /etc/crontabs/root || echo "$MIN * * * *	/usr/sbin/owm.lua" >> /etc/crontabs/root
+crontag -l | grep -q "owm.lua" || echo "$MIN * * * *	test -e /usr/sbin/owm.lua && /usr/sbin/owm.lua" >> /etc/crontabs/root
```
booo: This should be "crontab" not "crontag".

SvenRoederer:
superseded by PR #69

PR #65, opened 2015-12-31
[ffwizard] correcting dhcp-limit

SvenRoederer:
"limit"-option has to be 3 addresses less than the network-size
(network-, broadcast-, router-address)


SvenRoederer:
tested the original-patch (substract 2 addresses) on a real router and the code was ok. so the updates limit sould not break the code eighter.

@booo think we should push this to master

booo:
I will merge this soon. I will add a migration step too that I have prepared but not commited/pushed.

SvenRoederer:
sure, was just some feedback I'd like to give. I assume you need some recreation time after 32C3

PR #66, opened 2015-12-31
[ffwizard] do not enable VPN03, as this is controlled by hotplug-script

SvenRoederer:
In aacb343 setting the default-config of VPN03 was removed from
"freifunk-berlin-openvpn-files/uci-defaults/freifunk-berlin-openvpn" and is
incomplete by default. The OpenVPN-deamon wil be controlled by hotplugd, but
only when not enabled in UCI.
So don't enable it by default.


SvenRoederer:
Ist es denn eine gute Idee, so eine unvollständige uci-config zu haben? gibt das nicht Probleme in der Luci-gui?

sarumpaet:
Thanks for spotting that. How about just not touching OpenVPN config in the Assistent? Defaults are set by migration/the defaults packages anyways, so there's no reason to change that in the Assistent, and we would be overwriting possible manual changes by the user there.

UCI OpenVPN config: I'd be surprised if there were problems atm, but we can also remove defaults for that completely.

PR #67, opened 2016-01-02
Fix/dhcp start limit

SvenRoederer:
- fix migration; verified to work on updates from 0.1.2 to 0.2.0 
- be more verbose

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -125,13 +125,14 @@ fix_dhcp_start_limit() {
       local netmask
       local prefix
       # get network-length
-      netmask="$(uci get network.dhcp.netmask)"
-      # use ipcalc.sh to and get prefix-length only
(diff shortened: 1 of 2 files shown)


booo:
I don't think this is a good solution. We can't assume that the network interface has a netmask option.

SvenRoederer:
Why we can't assume?
- normal users flash their router to kathleen and end up in the wizard, which sets the correct DHCP-limit and also netmask
- normal users had setup their routers by the wizard, when they upgrade, we can rely on a correct net netmask
- advanced users should take care of the correct setup and change the limit by themself

Maybe we should check for a netmask and in case it's missing we skip the correction of the DHCP-limit,  as the user seems to have modifiied the uci-settings for some reason

SvenRoederer:
- added test for existing netmask
- be more verbose
- disable error-messages when uci is not finding the requested config-parameter
- tested it on a upgrade from kathleen 0.1.2 to 0.2.0-alpha

booo:
I tested the code for the wizard and it seems to work. @SvenRoederer successfully tested the upgrade. I think we can merge this.

PR #68, opened 2016-01-03
[ffwizard] also remove DHCP config for removed lan-interface

SvenRoederer:
checked, that it works as expected
think we don'T need to implement a migration step here, as it's not a functional thing


SvenRoederer:
checked, that it works as expected
think we don'T need to implement a migration step here, as it's not a functional thing

booo:
Once upon a time I was to lazy to implement this. Not sure why. Looks good.

Please try to explain what you do in the commit body too. This may seem like additional work but helps us understand changes later on in the process.

PR #69, opened 2016-02-14
Fix/correcting owm deps

SvenRoederer:
- this puts the "owm-defaults" file for setting up the cron-job to the owm-cmd package, where the called binary is inside
- add some checks for binary in cron-job
- rely on "crontab"-util to modify crontab (not "crond restart" required and saves some bytes on flash
  - tested this for empty and existing crontabs


booo:
Looks good to me. Any objections? If not I will merge this later this week.

PR #70, opened 2016-02-14
do not update OWM without latitude/longitude defined

SvenRoederer:
when the latitude/longitude are not defined for the node, /usr/sbin/owm.lua tries to update the map anyway. such updates get refused by OWM-server 
> fail   Doc  Statuscode: 403 http://api.openwifimap.net//update_node/Ahof-frieden03.olsr ({"error":"forbidden","reason":"Document must have a longitude"}
)

so prevent script from running without the informations, which also saves a lot of processing time
<pre>
# current code
root@Ahof-frieden03:~# time /tmp/owm.lua
fail   Doc  Statuscode: 403 http://api.openwifimap.net//update_node/Ahof-frieden03.olsr ({"error":"forbidden","reason":"Document must have a longitude"}
)
real    0m 4.10s
user    0m 0.47s
sys     0m 0.13s

# new code
root@Ahof-frieden03:~# time /tmp/owm.lua 
real    0m 0.26s
user    0m 0.23s
sys     0m 0.03s
</pre>

code works on my node

### DIFF
--- a/utils/luci-app-owm/files/owm.lua
+++ b/utils/luci-app-owm/files/owm.lua
@@ -53,7 +53,7 @@ function unlock()
 	os.execute("rm -f "..lockfile)
 end
 
--- a location for the node is required to run
+-- location for the node is required
 if next(owm.get_position())==nil then
 	if arg[1]=="--dry-run" then
 		print("no latitude/longitude specified for node.")
@@ -64,7 +64,7 @@ end
 
 lock()
 
-uci:foreach("system", "system", function(s) --owm
+uci:foreach("system", "system", function(s)
 	hostname = s.hostname
 end)
 
--- a/utils/luci-app-owm/files/owm.lua
+++ b/utils/luci-app-owm/files/owm.lua
@@ -54,7 +54,7 @@ function unlock()
 end
 
 -- location for the node is required
-if next(owm.get_position())==nil then
+if owm.get_position()==nil then
 	if arg[1]=="--dry-run" then
 		print("no latitude/longitude specified for node.")
 	end
--- a/utils/luci-app-owm/files/owm.lua
+++ b/utils/luci-app-owm/files/owm.lua
@@ -53,6 +53,15 @@ function unlock()
 	os.execute("rm -f "..lockfile)
 end
 
+-- a location for the node is required to run
+if next(owm.get_position())==nil then
+	if arg[1]=="--dry-run" then
+		print("no latitude/longitude specified for node.")
+	end
+	unlock()
+	return
+end
+
 lock()
 
 uci:foreach("system", "system", function(s) --owm
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -211,7 +211,11 @@ function get_position()
 		position['latitude'] = tonumber(s.latitude)
 		position['longitude'] = tonumber(s.longitude)
 	end)
-	return position
+	if (position['latitude'] and  position['longitude']) then
+		return position
+	else
+		return nil
+	end
 end
 
 function get()
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -207,9 +207,9 @@ end
 
 function get_position()
 	local position = {}
-	uci:foreach("system", "system", function(s) --owm
-		position['latitude'] = tonumber(s.latitude) --owm
-		position['longitude'] = tonumber(s.longitude) --owm
+	uci:foreach("system", "system", function(s)
+		position['latitude'] = tonumber(s.latitude)
+		position['longitude'] = tonumber(s.longitude)
 	end)
 	return position
 end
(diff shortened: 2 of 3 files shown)


utils/luci-app-owm/luasrc/owm.lua
```diff
@@ -205,6 +205,15 @@ function showmac(mac)
 	return mac
 end
 
+function get_position()
+	local position = {}
+	uci:foreach("system", "system", function(s) --owm
```
booo: Why do we use `uci:foreach` here? I would use `uci:get("system", "system", "latitude")`.

utils/luci-app-owm/files/owm.lua
```diff
@@ -53,6 +53,15 @@ function unlock()
 	os.execute("rm -f "..lockfile)
 end
 
+-- a location for the node is required to run
+if next(owm.get_position())==nil then
```
booo: Maybe `owm.get_position()` should just return `nil` if we can't get the position from the configuration file? This `next` call is a bit irritating.

utils/luci-app-owm/luasrc/owm.lua
```diff
@@ -205,6 +205,15 @@ function showmac(mac)
...
+		position['latitude'] = tonumber(s.latitude) --owm
```
booo: We should remove the comments...

utils/luci-app-owm/luasrc/owm.lua
SvenRoederer: I had seen this also and was asking for short, but also think it's not usefull here

utils/luci-app-owm/files/owm.lua
SvenRoederer: an empty table is not NIL, so this "next($tablevar)" can detect this. get_position()==nil will always be true.

this kind of test only needs 6bytes on disc, checking the values of long & lat and return nil or the data as table will cost more flash-space

booo: I don't care about the extra flash space this takes. I care about readable code. `next(owm.get_position())==nil` is not quite intuitive.

utils/luci-app-owm/luasrc/owm.lua
SvenRoederer: cleaned / moved them back to original positions. but there are many of these "--owm" comments.

looks like they are marking the datafields of the OWM-json
@stargieg what do you think

utils/luci-app-owm/luasrc/owm.lua
SvenRoederer: think that's the reason
> system.@system[0]=system
system.@system[0].cronloglevel='0'
system.@system[0].zonename='Europe/Berlin'
system.@system[0].timezone='CET-1CEST,M3.5.0,M10.5.0/3'
system.@system[0].hostname='Ahof-frieden03'
system.@system[0].version='0.2.0-alpha'
system.@system[0].location='test'
system.@system[0].latitude='52.436757800213844'
system.@system[0].longitude='13.546518087387085'

utils/luci-app-owm/luasrc/owm.lua
stargieg: -----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

Hi

Am 14.02.16 um 18:03 schrieb Sven Roederer:
>> @@ -205,6 +205,15 @@ function showmac(mac)
>>  	return mac
>>  end
>>  
>> +function get_position()
>> +	local position = {}
>> +	uci:foreach("system", "system", function(s) --owm
>> +		position['latitude'] = tonumber(s.latitude) --owm
> 
> cleaned / moved them back to original positions. but there are many of these "--owm" comments.
> 
> looks like they are marking the datafields of the OWM-json
> @stargieg what do you think
Yes. This fields are necessary for owm db

-----BEGIN PGP SIGNATURE-----
Version: GnuPG/MacGPG2 v2
Comment: GPGTools - http://gpgtools.org

iEYEARECAAYFAlbAvxgACgkQr9m0OkMZoqTB7ACgpizUcmtWMHLLQx4kw8dOUQ6j
MB4An1+tDOtzJktBHOewUU3w2gP/rNXI
=7U1L
-----END PGP SIGNATURE-----

SvenRoederer: So any preference to keep them or to remove these "--owm" comments?

stargieg: -----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA1

Am 14.02.16 um 18:58 schrieb Sven Roederer:
>> @@ -205,6 +205,15 @@ function showmac(mac)
>>  	return mac
>>  end
>>  
>> +function get_position()
>> +	local position = {}
>> +	uci:foreach("system", "system", function(s) --owm
>> +		position['latitude'] = tonumber(s.latitude) --owm
> 
> So any preference to keep them or to remove these "--owm" comments?
Remove and make a hint where is data structure documented
https://github.com/freifunk/openwifimap-api#standard-fields

-----BEGIN PGP SIGNATURE-----
Version: GnuPG/MacGPG2 v2
Comment: GPGTools - http://gpgtools.org

iEYEARECAAYFAlbAwYcACgkQr9m0OkMZoqSd8wCdErBuRriia5i0DXCQPG5CQved
KHUAnRm978WZcr37vy5OxSlNG9NKogL0
=pqOD
-----END PGP SIGNATURE-----

PR #71, opened 2016-02-23
remove "latlon" from uci: system.system (resend)

SvenRoederer:
- this removes the uci-setting "system.system.latlon"
- "latlon" seems to be only historic, also by checking the whole src-tree (relevant for Freifunk) there seems to be no active use of the parameter; only found some fallback in case "system.system.latitude" and "system.system.longitude" are not defined
- code seems to work fine on my systems
- this closes https://github.com/freifunk-berlin/firmware/issues/307
- is a resubmit of PR #61 

- is in public beta in all unstable-firmware images of https://github.com/freifunk-berlin/firmware/tree/unstable_SAm0815 without any known problems

### DIFF
--- a/utils/luci-app-ffwizard-berlin/Makefile
+++ b/utils/luci-app-ffwizard-berlin/Makefile
@@ -3,7 +3,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-ffwizard-berlin
-PKG_VERSION:=0.0.1
+PKG_VERSION:=0.0.2
 PKG_RELEASE:=1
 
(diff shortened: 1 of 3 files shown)


SvenRoederer:
@boo as discussed, the case to convert a "foreign" OpenWRT-flavor to "freifunk Berlin" openWRT is a bit esotheric ...
this was the thing I had in mind during renaming of "system.version" to "system.kathleen_version"

maybe it'S better to revert this and stay with current setting

booo:
Yes, looks good. Sorry, for the slow progress.

PR #72, opened 2016-03-24
[openvpn] Fix openvpn hotplug trying to start with invalid uci config

SvenRoederer:
- catches https://github.com/freifunk-berlin/firmware/issues/338 and gives at least some logmessage
- it's only an informative message, but helps the user to find the source of the problem


SvenRoederer:
@sarumpaet did some rework on the hotplug proposal which I'll  send in an other PR

PR #73, opened 2016-03-24
extend "openvpn hotplug" to run via procd

SvenRoederer:
- in recent OpenWRT CC the OpenVPN is run via procd
- using this daemon seems more reliable in case if OpenVPN-crashes
- using the default init-script also allows using default UCI-config
- also implements the nice idea binding to the WAN-interface to prevent OpenVPN over freifunk-mesh, and gets rid of using firewall-rules for blocking this traffic

also takes care of https://github.com/freifunk-berlin/firmware/issues/338

### DIFF
--- a/utils/freifunk-berlin-openvpn-files/openvpn/60-ffopenvpn
+++ b/utils/freifunk-berlin-openvpn-files/openvpn/60-ffopenvpn
@@ -20,15 +20,17 @@ config_get uci_vpn_enabled ffvpn enabled
 
 if [ "$ACTION" = ifup ]; then
   logger -t ff-userlog "WAN interface is up, starting OpenVPN"
+  # Make sure OpenVPN connects only through WAN: set "local" option with WAN IP
+  local wanip
+  network_get_ipaddr wanip "wan"
   if [ "$uci_vpn_enabled" = 1 ]; then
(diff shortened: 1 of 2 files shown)


SvenRoederer:
this further enhances https://github.com/freifunk-berlin/firmware-packages/pull/62

booo:
I'm fine with the minimal change of the configuration by the hotplug script. Still I don't like the hidden configuration in the script and I don't want the openvpn process started if I disable the ffvpn configuration. I suggest that we remove the configuration from the script and don't start openvpn if it's disabled.

sarumpaet:
The idea of putting the VPN03 config in the script is that a) no user needs to/should edit the VPN03 config ever really b) if someone wants to use a custom config, he can use UCI config, and c) moving the VPN03 config to the script removes several dozens of code/config and one migration. Also, I think this goes in the spirit of "Kathleen is for Freifunk Berlin, not some poor try at OpenWrt++".

The ffvpn UCI config should not really be created at all.

@booo The OpenVPN process doesn't get started automatically unless the user chose to, see `[ "$sharenet" = 1 ] || exit`.

SvenRoederer:
the big pro for using UCI and /etc/init.d is the seamless integration of procd for keep OpenVPN running.

PR #74, opened 2016-03-24
Fixes running migration even on fresh install

SvenRoederer:
found that on a fresh install the migration-script gets started, when working on 
- https://github.com/freifunk-berlin/firmware-packages/pull/72
- https://github.com/freifunk-berlin/firmware-packages/pull/73

This is caused by the fact that UCI "system.version" in not defined and the default "0.0.0" is choosen.

Running migraition on fresh installs makes no sense and has the chance to break some settings.

This idea here, tries to find the difference between a fresh install (of a >= kathleen 0.2.0) firmware and running on a old firmware by looking for files older than the first kathleen-release. See "ctime" of file /tmp/timestamp for this (Line34).
Only the "basic" Semver numbers are taken into account when deciding on migration-steps. The current code as some problems with the version-string when a "special-version" (-alpha+xxx) is suffixed.

by this chance UCI "system.version" gets renamed to "system.kathleen_version".

### DIFF
--- a/utils/freifunk-berlin-migration/Makefile
+++ b/utils/freifunk-berlin-migration/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-migration
-PKG_VERSION:=0.0.0
+PKG_VERSION:=0.1.0
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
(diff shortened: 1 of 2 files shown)


SvenRoederer:
as discussed with @booo , the case to convert a "foreign" OpenWRT-flavor to "freifunk Berlin" openWRT is a bit esotheric ...
this was the thing I had in mind during renaming of "system.version" to "system.kathleen_version"

So I reverted it to system.version again

SvenRoederer:
I ran some test on 2 or 3 routers and the log in "/root/migration.log" let me feel it's working.

SvenRoederer:
just tested this with the latest text-changes on a fresh-install and an upgrade from 0.2.0-alpha and all looks fine.
was tested with "unstable/ar71xx/876/"; https://github.com/SvenRoederer/freifunk-berlin-firmware-packages/commit/00dff35acd5c03ec0ced1c2260c38aa02730bcb5

PR #75, opened 2016-04-03
migration: add migration for removal of dhcp.lan config

SvenRoederer:
in addition to the https://github.com/freifunk-berlin/firmware-packages/pull/68 which became https://github.com/freifunk-berlin/firmware-packages/commit/cd4d319499338ff5b088a441acaed1847ea34d5c

we changed this th ffwizard, but forgot about a migration for this


booo:
I'm not sure if adding a migration step is a good idea for people which use the backbone image and configured a custom lan section.

sarumpaet:
Backbone doesn't run migrations, see https://github.com/freifunk-berlin/firmware/issues/231

booo:
Ok. I think it's fine to merge this.

PR #76, opened 2016-04-05
[ffwizard] remove freifunk-watchdog configuration from assistent

booo:
Functionality provided by the freifunk-watchdog is now provided by procd. Other
functionality like a reboot at high load is nothing we need.

For a longer discussion please take a look at:

https://github.com/freifunk-berlin/firmware/issues/52

I think this need still needs a migration step.

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -118,7 +118,7 @@ fix_qos_interface() {
 }
 
 remove_freifunk_watchdog_from_crontab() {
-  crontab -l | grep "/usr/sbin/ffwatchd" && crontab -l | grep -v "/usr/sbin/ffwatchd" | crontab -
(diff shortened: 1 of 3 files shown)


booo:
I think we can use the crazy code from here: https://github.com/openwrt/luci/blob/master/contrib/package/freifunk-watchdog/files/freifunk-watchdog.init#L19

SvenRoederer:
We can even use something similar to this https://github.com/freifunk-berlin/firmware-packages/blob/c2429d09e4c1bc4ef1514af88dea766ab9b79381/utils/luci-app-owm/files/owm-defaults#L5

> crontab -l | grep "/usr/sbin/ffwatchd" && crontab -l | grep -v "/usr/sbin/ffwatchd" | crontab -

booo:
What do you think? I tested the command on my router but not the migration script. Good to merge?

SvenRoederer:
> crontab -l | grep -qv "/usr/sbin/ffwatchd" | crontab -

should do the same and is more easy to read

sarumpaet:
see #62

PR #77, opened 2016-04-18
fix migration on fresh install

SvenRoederer:
see https://github.com/freifunk-berlin/firmware/issues/340 (when setting up a router from scratch, the migration-script gets started)

this fixes this by assuming any install without UCI-setting system.version defined is a fresh install and there is no need for Migration.
There is also some kind of legacy-support to migrate "old freifunk"-systems (pre-kathleen) by checking for files created before kathleen was released.
When UCI-setting system.version is defined is acts straight forward.


booo:
Merged the version that generates a migration log in /root/migrate.log.

PR #80, opened 2016-04-25
luci-mod-freifunk-ui

sarumpaet:
luci-mod-freifunk-ui: just the UI parts of luci-mod-freifunk including our FFB patches and without freifunk-firewall dependency.


SvenRoederer:
the patches references in the PR-description is https://github.com/freifunk-berlin/firmware-packages/commit/77eaab3cda479dd2b6baf432a8c3619a11a5f57f and probably others

PR #81, opened 2016-04-26
Fix/dhcp start limit (updated to apply on current master)

SvenRoederer:
This was totally missed during the firmware-hacking event ....

updated the branch to apply against current master

as before this code did not show any problems during public-beta-testing in "unstable_SAm0815"-builds of firmware


booo:
Where does this alter the dhcp configuration? I'm confused...

SvenRoederer:
it'S just a patch for the branch, to make it apply to the current master.
As the "basic idea" is still the same, you only see the diff to make it apply.

after accepting the PR the fix/dhcp_start can be applied directly

PR #82, opened 2016-04-26
Fix/remove ff watchdog missing migration

SvenRoederer:
we missed the migration for removal of freifunk-watchdog, to clean up the crontab

* https://github.com/freifunk-berlin/firmware-packages/commit/7a6a120385571e7772acbe87fd8d9b0c234cb5a8
* https://github.com/freifunk-berlin/firmware/commit/39ecc22d0cf53cbf1a2e853984fda956b2d5ae21

apply this after PR #81


booo:
As far as I understand the situation we should just merge this request. Right?

sarumpaet:
The crontab ffwatchd thing is already in openvpn_ffvpn_hotplug.

SvenRoederer:
ok, missed this  - did not expect it in the openvpn-hotplug migration :-)

booo:
Maybe we should move this in a quick change into another function?

SvenRoederer:
it's cosmetic, but I felt into this trap already ... would look more nice

sarumpaet:
@SvenRoederer Actually I talked about exactly that line with you several times yesterday ;)

SvenRoederer:
@sarumpaet hmm, I'm getting old?

PR #83, opened 2016-05-06
[ffwizard] never try to auto-login if password has been set

sarumpaet:
If a user sets a password but doesn't finish the set up process in
the wizard, the initial ffwizard auto-login redirect fails. This
adds a flag so that no auto-login happens if a password has been set.
Fixes https://github.com/freifunk-berlin/firmware/issues/208

### DIFF
--- a/utils/luci-mod-freifunk-ui/luasrc/view/freifunk/index.htm
+++ b/utils/luci-mod-freifunk-ui/luasrc/view/freifunk/index.htm
@@ -22,9 +22,11 @@ local webAppRoot = http.getenv("PATH_INFO") == nil
 local notRunBefore = not uci.get("ffwizard", "settings", "runbefore")
 local notPasswordSet = not uci.get("ffwizard", "settings", "passwordset")
 local wizardInstalled = fs.access("/usr/lib/lua/luci/controller/assistent/assistent.lua")
-if (webAppRoot and notRunBefore and notPasswordSet and wizardInstalled) then
+if (webAppRoot and notRunBefore and wizardInstalled) then
 	local url = luci.dispatcher.build_url("admin/freifunk/assistent")
-	url = url .. "?luci_username=root&luci_password="
+	if (notPasswordSet) then
+		url = url .. "?luci_username=root&luci_password="
+	end
 	http.redirect(url)
 end
 %>
--- a/utils/luci-mod-freifunk-ui/luasrc/view/freifunk/index.htm
+++ b/utils/luci-mod-freifunk-ui/luasrc/view/freifunk/index.htm
@@ -16,12 +16,13 @@ local disp = require "luci.dispatcher"
 local ipkg = require "luci.model.ipkg"
 
 -- only redirect if assistent is installed and no root password is set
--- we use isfile because the view is run as root and we can't use e.g.
--- ipkg.installed or checkpasswd
+-- we use isfile because the view is not run as root and we can't use
+-- e.g. ipkg.installed or checkpasswd
 local webAppRoot = http.getenv("PATH_INFO") == nil
 local notRunBefore = not uci.get("ffwizard", "settings", "runbefore")
+local notPasswordSet = not uci.get("ffwizard", "settings", "passwordset")
 local wizardInstalled = fs.access("/usr/lib/lua/luci/controller/assistent/assistent.lua")
-if (webAppRoot and notRunBefore and wizardInstalled) then
+if (webAppRoot and notRunBefore and notPasswordSet and wizardInstalled) then
 	local url = luci.dispatcher.build_url("admin/freifunk/assistent")
 	url = url .. "?luci_username=root&luci_password="
 	http.redirect(url)
--- a/utils/luci-app-ffwizard-berlin/luasrc/controller/assistent/assistent.lua
+++ b/utils/luci-app-ffwizard-berlin/luasrc/controller/assistent/assistent.lua
@@ -33,7 +33,7 @@ function prepare()
     uci:save("ffwizard")
     uci:commit("ffwizard")
   end
-  if not uci:get("ffwizard","settings","runbefore") then
+  if not uci:get("ffwizard","settings","runbefore") and not uci:get("ffwizard", "settings", "passwordset") then
     luci.http.redirect(luci.dispatcher.build_url("admin/freifunk/assistent/changePassword"))
   else
     luci.http.redirect(luci.dispatcher.build_url("admin/freifunk/assistent/generalInfo"))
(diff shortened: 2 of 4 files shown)


lynxis:
LGTM but I don't usually code lua. merge if you like.

SvenRoederer:
Why not checking if the password is set in /etc/password (/etc/shadow)?
Having this in flag in UCI ffwizard will not work if the user changed the password via console or regular web-UI.

sarumpaet:
Look at the comment in index.htm: The view isn't run as root and does not have access to /etc/shadow.
The approach does not cover all the possible cases, but it DOES cover all the REPORTED cases I think, plus it doesn't require another ugly OpenWrt patch. I'd say if someone wants to fix it upstream, go ahead, until then we should just merge this, it'll be good enough.

SvenRoederer:
@sarumpaet : btw. when not running as root, as which user the script is running ? Just for my curiosity

SvenRoederer:
what about ubus + rpcd-plugin:
```
root@sam0815-test:~# cat /usr/libexec/rpcd/freifunk_berlin 
#!/bin/sh

# taken from https://wiki.openwrt.org/doc/techref/rpcd
case "$1" in
        list)
                echo '{ "root_has_pwd": {} }'
        ;;
        call)
                case "$2" in
                        root_has_pwd)
                                local USER=root
                                # taken from http://serverfault.com/a/240961
                                grep ${USER} /etc/shadow | awk -F":" '{ if ($2 == "!" || $2 == "*") \
                                                print "\{\"password_is_set\": \"no\"\}"; \
                                        else \
                                                print "\{\"password_is_set\": \"yes\"\}"\
                                        }'
                        ;;
                esac
        ;;
esac
```
This can creates an object in ubus which can be queried:
> root@sam0815-test:~# ubus call -S -v freifunk_berlin root_has_pwd 
{"password_is_set":"yes"}

And then via libubus to lua:
```
 root@sam0815-test:~# cat /root/check_for_rootpwd.lua 
require "ubus"

local conn = ubus.connect()
if not conn then
    error("Failed to connect to ubusd")
end

local status = conn:call("freifunk_berlin", "root_has_pwd", {})
for k, v in pairs(status) do
    print("key=" .. k .. " value=" .. tostring(v))
end

conn:close()
```

gives 
>root@sam0815-test:~# lua check_for_rootpwd.lua 
key=password_is_set value=yes


This would work no matter which way the password was set.

sarumpaet:
Wow. This adds two scripts and quite a bit of complexity and brittleness (how long until some OpenWrt update will break this?) to cover a case that doesn't matter in practice and could be easily fixed with about one line in upstream (see the very first comment in https://github.com/freifunk-berlin/firmware/issues/208 ). I applaud the effort, but how about directing this energy into something like https://github.com/freifunk-berlin/firmware/issues/292 ? :)

SvenRoederer:
only one plugin for rpcd and a change of the code in the ffwizard, that will be changed anyway.

I don't think that the rpcd will be canceled in the near future, so not expecting it to becom incompatible.

booo:
I favor a solution that fixes the problem once and for all. Either we try to implement the proposal from https://github.com/freifunk-berlin/firmware/issues/208 or we take the rpcd solution.

Anyway I think we should solve this in the next release. It's anyway broken since a few releases and we should really push a new release now!

SvenRoederer:
+1 for keeping this for the next release

bobster-galore:
I think we should decide regarding our resources if this is easily finished in one way or another we should include it in 0.2.0. If not we can postpone it to a later release. cite "to cover a case that doesn't matter in practice"
To cut it short if there is no surplus of bored willing programmers we should keep it in mind for the next release.

sarumpaet:
I'd just merge the PR as-is, then replace it by a "complete" solution later if need be.

lynxis:
This is just a small fix, I'm going to merge it.
@SvenRoederer if you want to do it that way, it would also introduce a dependency rpcd which is not yet part of firmware. But might be in the future.

SvenRoederer:
@lynxis There is no direct dependency on rpcd, but luci-base has one. So it's included already anyways.

lynxis:
@SvenRoederer ok, I see. I only looked into our old release, which did not contain rpcd. BTW. The script also contain a bug when the password string is empty.

This PR is just a quick and simple fix for a problem which will be solved by another wizard. I don't see big problems with this PR.

SvenRoederer:
https://github.com/freifunk-berlin/firmware-packages/pull/83#issuecomment-232204595 also fail on "LuCI not running as root", as the code can not access the UBus-socket, which is mode 600 for user ang group "root".

Going this way will require to go via "uhttp-mod-ubus" and implementing some ACL on UBus/RPC for this.

sarumpaet:
Let's merge this? Just tested again, works.

In case the user set the password via SSH or something before running the wizard, he can still log in just by deleting the URL parameters (i.e., by going to `/cgi-bin/luci/admin/freifunk/assistent`) or by clicking "Administration" to the bottom right.

SvenRoederer:
do we need a migration for this (uci:set("ffwizard", "settings", "passwordset", "true")) ?

if so maybe inverting the logic might be an option. 
to change to option to "passwordunset" and having it "true" after initial flash, so current setups are not affected automatically

sarumpaet:
No migration is necessary. For routers with Kathleen 0.1.2 config, runbefore is set to "true", so both IFs will correctly bail out on the "if not runbefore..." already.

utils/luci-mod-freifunk-ui/luasrc/view/freifunk/index.htm
```diff
@@ -16,12 +16,13 @@
 local ipkg = require "luci.model.ipkg"
 
 -- only redirect if assistent is installed and no root password is set
--- we use isfile because the view is run as root and we can't use e.g.
--- ipkg.installed or checkpasswd
+-- we use isfile because the view is not run as root and we can't use
```
SvenRoederer: can you put this in a separate commit to get this more easy merged to upstream?

sarumpaet: There is no upstream for this; the "redirect to assistent" thing is Berlin specific.

SvenRoederer: i should have checked this before ... did not believe that we are so much apart from upstream

booo:
@sarumpaet Can you review the latest changes and merge this if you are happy with them?

To me the final solution looks good.

sarumpaet:
Fixed in https://github.com/freifunk-berlin/firmware-packages/pull/89

PR #84, opened 2016-05-06
[ffwizard] add altitude

sarumpaet:


booo:
What was the reasoning for adding an altitude field? Can you add the reasoning to the commit?

I would prefer to not add another field to the wizard especially if we don't have a direct use case for this data.

sarumpaet:
Antenna documentation (see https://github.com/freifunk-berlin/firmware-packages/tree/owm_antennas ) is not complete without antenna altitude, and obviously it's quite important to document whether, e.g., an omni antenna is on street level or on a 10th floor.

Anyways, this is not important for 0.2.0.

bobster-galore:
I see the documentary relevance, is it somehow working and tested or should we postpone it to the next release?

sarumpaet:
To be complete, this need modifications to OWM, and possibly a server side component (to calculate height/sea level from height/ground), but let's merge this.

PR #85, opened 2016-07-30
snmp-templates: add template to query Ubiquiti AirMax via SNMP

SvenRoederer:
example at: http://monitor.berlin.freifunk.net/host.php?h=rhnk-adlershof&p=snmp


SvenRoederer:
any concerns with putting this into Kathleen 0.2.0?

There is some user demand, see https://lists.berlin.freifunk.net/pipermail/berlin/2016-October/034313.html ++

SvenRoederer:
Timeout - w/o any negative feedback

PR #86, opened 2016-08-03
Fix/owm stats for kathleen 0.2.0

SvenRoederer:
This code implements a compatible json-dataset to send Router-model and Chipset-type as in kathleen 0.1.2. 
So the statistics-wegpage will work as it is (see issue https://github.com/freifunk-berlin/firmware/issues/378#issuecomment-237082534).

You can copy the owm.lua to the router under /usr/lib/lua/luci to for direct testing

### DIFF
--- a/utils/luci-app-owm/Makefile
+++ b/utils/luci-app-owm/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-owm
-PKG_RELEASE:=0.4.11
+PKG_RELEASE:=0.4.12
 
 
 PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -24,6 +24,7 @@ local status = require "luci.tools.status"
 local json = require "luci.json"
 local netm = require "luci.model.network"
 local sysinfo = luci.util.ubus("system", "info") or { }
+local boardinfo = luci.util.ubus("system", "board") or { }
 local table = require "table"
 local nixio = require "nixio"
 --local neightbl = require "neightbl"
@@ -251,7 +252,7 @@ function get()
 	root.hostname = sys.hostname() --owm
 
 
-	root.hardware = "unknown" --owm TODO
+	root.hardware = boardinfo['system'] --owm
 	
 
 	root.firmware = {
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -39,6 +39,27 @@ local ipairs, os, pairs, next, type, tostring, tonumber, error, print =
 module "luci.owm"
 
 
+-- backported from LuCI 0.11 and adapted form berlin-stats
+--- Returns the system type (in a compatible way to LuCI 0.11)
+-- @return	String indicating this as an deprecated value
+--        	(instead of the Chipset-type)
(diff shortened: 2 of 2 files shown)


bobster-galore:
even if it's not compiled?

SvenRoederer:
yes, with the drawback of runtime-compilation on every run.

sarumpaet:
@SvenRoederer I don't quite see how the sysinfo is really "backported" from LuCI 0.11? The LuCI 0.11 sysinfo is here: https://github.com/openwrt/luci/blob/luci-0.11/libs/sys/luasrc/sys.lua#L162 - it's quite a bit different from your version - doesn't send trailing newlines, and includes memory stats. Not that I'd care about that in context of website stats, but it was you who said he wanted backwards compatibility, so why changing that function?

bobster-galore:
I think it's more a backport reduced to our needs. 
The owm-part is a snapshot at least an hour old. 

- It feeds the data into the map, where nothing about memory or usage is displayed.

- It feeds the statistics, where the relevant fields are router-model, hw-type (chipset) and fw-Revision.

The for now omitted data are also collected by collectd and displayed at monitor including a view over time. I see no need to collect the same data twice, when one isn't even used.

The trailing newline seems to be generated somewhere, without the 0.2.alpha-routers are listed seperately.

sarumpaet:
Probably I was just confused by Sven's "break with all old releases" in https://github.com/freifunk-berlin/firmware/issues/378#issuecomment-236997126 - I thought he wanted to have the OWM JSON the same way it was in Kathleen 0.1.2, whereas now it says "backport" yet it isn't.

Anyways.

@SvenRoederer 
* remove the trailing newline https://github.com/freifunk-berlin/firmware-packages/pull/86/files#diff-1190ed871db168151a64d3f9bd15b0ceR52
* remove the wrong ref to memory stats https://github.com/freifunk-berlin/firmware-packages/pull/86/files#diff-1190ed871db168151a64d3f9bd15b0ceR43
* fix the typo https://github.com/freifunk-berlin/firmware-packages/pull/86/files#diff-1190ed871db168151a64d3f9bd15b0ceR49 (it's "deprecated")

Then pls merge and thanks :)

bobster-galore:
We should clarify where the trailing newline comes from. @sarumpaet if you look at the json-data of kath-0.0.-1.2.-routers you will find that their model comes with a trailing newline (s. christophorus-n). If it's not in the owm.lua code, where does it come from?

This is why Sven added the newline, otherwise there will be the same model counted separate.

SvenRoederer:
I initial had not this "\n" on the ubus-output and the same router-model (TPL WR1043v1) were not aggregated (see https://github.com/freifunk-berlin/firmware/issues/378#issuecomment-237197004). 

As noted by @bobster-galore the newline (from LuCI sysinfo()) makes the difference.

sarumpaet:
Oh. I guess that was a bug in the old LuCI `model = luci.util.pcdata(fs.readfile("/tmp/sysinfo/model"))` then. Clearly the newline shouldn't be there (note the other cases that exclude it). I'm not fond of bug compatibility, especially as this will break for the "OR" cases in the new code (that will not include newlines). https://github.com/freifunk-berlin/berlin.freifunk.net/commit/7120a7b09f6b7766583220136edbe35dba79dea1 adds a trim() in the stats so it'll be fine in any case, and we need not do bug downwards compatiblity here.

bobster-galore:
trim is nice! Since it is always generated fresh there is no hazzle with old strings. (I hope)

but sad, I like the sound of "bug downwards compatibility"

SvenRoederer:
@sarumpaet I was also thinking of some trim() but JavaScript is not mine ...

PR #87, opened 2016-09-26
[ffwizard] add a note to check all settings after configuration

SvenRoederer:
Inform the user to check the settings done by the assistent. This is
recommended esp. for the WiFi-settings as some there are some routers
that might have incorrect defaults.

### DIFF
--- a/utils/luci-app-ffwizard-berlin/Makefile
+++ b/utils/luci-app-ffwizard-berlin/Makefile
@@ -3,7 +3,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-ffwizard-berlin
-PKG_VERSION:=0.0.3
+PKG_VERSION:=0.0.4
 PKG_RELEASE:=1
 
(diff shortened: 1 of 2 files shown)


SvenRoederer:
this is somehow related to https://github.com/freifunk-berlin/firmware/issues/381 to make the user aware to operate within legal limits

sarumpaet:
I really dislike such legalese. Also, the link given does not say anything about TX power. If we really want to add something like this, we should give proper advice (including *what* to check for, *how* to check it and *how* to fix it if broken) instead of effectively linking to BNetzA which is not helpful.

booo:
I agree with @sarumpaet and I think this will only create confusion.

@SvenRoederer if you have strong opinions on this please reopen this ticket.

SvenRoederer:
how do you think of a link to the Releasenote?

sarumpaet:
* we'd have to update that link every release
* so far we didn't really write proper release notes ever _really_
* the router/computer is offline when showing the last screen anyways.

PR #88, opened 2016-10-15
network-defaults: add workaround for too high txpower on some units

SvenRoederer:
this is a workaround for https://github.com/freifunk-berlin/firmware/issues/381.
Some units (esp. Nanostations M2) use a too high TX-power by default, caused by
upstream.
This script relies on iwinfo where antenna-gains can be defined and just
subtracts the gain from the allowed upper limit.
It only runs once per router, as of guard "2GHz_txpower"

The currently known router with antenna-gains other than "0" have only one
wifi-phy, so only one PHY needs to be checked. Even this interface is by OpenWRT-
default "wlan0" and per Freifunk-default "wlan0-dhcp-2" and "wlan0-adhoc-2"

### DIFF
--- a/utils/freifunk-berlin-network-defaults/Makefile
+++ b/utils/freifunk-berlin-network-defaults/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
-PKG_VERSION:=0.0.1
-PKG_RELEASE:=1
+PKG_VERSION:=0.0.2
+PKG_RELEASE:=2
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
@@ -13,7 +13,7 @@ define Package/freifunk-berlin-network-defaults
   CATEGORY:=freifunk-berlin
   TITLE:=Freifunk Berlin network default configuration
   URL:=http://github.com/freifunk-berlin/packages_berlin
-  DEPENDS+= +freifunk-berlin-lib-guard
+  DEPENDS+= +freifunk-berlin-lib-guard +iwinfo
 endef
 
 define Package/freifunk-berlin-network-defaults/description
--- a/utils/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-correct-nsm2-txpower
+++ b/utils/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-correct-nsm2-txpower
@@ -1,0 +1,31 @@
+#!/bin/ash
+#
+# For Ubiquiti the default TX-power on OpenWrt is above regulatory limits.
+# see https://github.com/freifunk-berlin/firmware/issues/381
+#
+
+iwinfo|grep -q 'NanoStation M2\|NanoStation Loco M2' || exit 0
+
+. /lib/functions/guard.sh
+guard "NSm2_txpower"
+
+set_default_txpower() {
+ echo "setting txpower value to $1 dBm"
+ uci set wireless.radio0.txpower=$1
+ uci commit wireless
+}
+
+MAX_TX_2G=20
+# get "TX offset" of 1st interface, if this is unknown set it to 0dB
+# then only return the numerical value
+ANT_GAIN=$(iwinfo |grep -m 1 "TX power offset:" |sed -e "s/unknown/0 dB/" |tr -dc '0-9')
+NEW_TX=$(echo $(($MAX_TX_2G - $ANT_GAIN)))
+
+CURR_TX=$(uci -q get wireless.radio0.txpower)
+# check if txpower is defined in config
+if [ $? -ne 0 ]; then
+  echo -n "txpower not defined - "
+  set_default_txpower $NEW_TX
+elif [ $CURR_TX -gt $NEW_TX ]; then
+  set_default_txpower $NEW_TX
+fi


utils/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-default-txpower
```diff
@@ -0,0 +1,33 @@
+#!/bin/ash
+#
+# On Ubiquiti it has been seen that the default TX-power on OpenWRT
+# is at setting of Hardware, which is above regulatory limits
+# see https://github.com/freifunk-berlin/firmware/issues/381
+#
+
+. /lib/functions/guard.sh
+guard "2GHz_txpower"
```
sarumpaet: "2GHz_txpower" sounds generic, any reason against making this something like "ubnt_nsm_24ghz_txpower"?

SvenRoederer: the "iwinfo"-method adjusts against the defined offsets in http://git.openwrt.org/?p=project/iwinfo.git;a=blob;f=hardware.txt;h=81b31cfcbca1d00cfabbdb6768ab9e0f97a8cc89;hb=d22c64c2eb6db3db6aaac5ae270acd63ed088a20.
So this works on more hardware than just the NanoStations, but infact atm only works for Ubiquiti. But this can change by the next releases of iwinfo.
So "2GHz_txpower" is currently maybe a bit wide, but "ubnt_nsm_24ghz_txpower" is a bit to specific.
If we want to narrow down we can do something like "ubnt_24ghz_txpower", which reflects the current hardware-list

booo:
I suggest that we move this into a script with the name "*-wireless-defaults".

SvenRoederer:
I also had the idea of moving the call of "guard()" into the "set_default_txpower()"-function.

So this hack will be called on every router on a fresh install or upgrade till its was once actively setting a new default. So we can ensue that also boards currently not defined in iwinfo will be set to the correct limits after they have been added upstream.
Or we leave it as a hack and hope for a fix on the next release.

sarumpaet:
Calling guard() _after_ changing the UCI config doesn't make sense. It should be called before.

> On Ubiquiti it has been seen that the default TX-power on OpenWRT is at setting of Hardware,
> which is above regulatory limits

I'm sorry - what is this supposed to mean? Is this a typo?

> We use the iwinfo-tool to lookup the antenna-gain

How about something like "We read the TX power offset reported by iwinfo and assume that is the antenna gain".

> and might be able  to set the value on any supported board.

"Might"?


I'd really rather have this as "IF we're on a NSM2 AND guard not set AND wireless.radio0.txpower not set THEN set txpower". The current version might be too broad and break stuff, looks brittle (iwinfo scraping), and seems difficult to test (why does it run on all routers instead of only the one it's supposed to fix?). This doesn't appear like the thing to add a few days before a release.

SvenRoederer:
@sarumpaet I started with this script as you mentioned above, but came to the point where I had to distinguish between NSM2 and NSLocoM2 (as of different antenna-gain).
I only found "iwinfo" seems to be able to detect this. As you remember NSM2 / M5 are even the same board and this is also true for Loco / non-Loco. 

and from using "iwinfo" to detect the board and do something hardcoded to completly rely on "iwinfo" is not such a long way.

tested this on multiple NSMs (M2, M5, Locos) and also other boards - for good.

sarumpaet:
So why not do
```
guard
ubus -v call system board|grep -q "Ubiquiti Nanostation" || exit
(txpower magic)
```

Re boards, there's also quite a few other devices (e.g., GL.iNet, Nexx, D-Link).

SvenRoederer:
the other devices will fall into the "generic mac80211" section of iwinfo as they are not defined and so so they act like the silicon I tested on (no ant-gain defined --> no change)

As the suggested board for NSLocoM2 is the bullet image you have to check for this also and with iwinfo later the lookup of the real board again.
My approach has only one location for having the board-data, not two separate as you suggest.

sarumpaet:
Again: The effect of this hack should be to (only) set txpower on NSM2s. At the moment, for example, if for whatever reason I have wireless.radio0.txpower=30 on a router of mine (say a TP-Link), this script will tamper with that. This script isn't supposed to do that, it's as simple as that.

SvenRoederer:
This script is not intended to be restricted to NSM2 only. It's, as the name shows, to set the max. tx-power allowed in ETSI-region. The NSMs are just a most prominent example, which presented this issue. The issue causing this hack is present for all Hardware with fix Ant-Gain, so I targeted it to all boards, that are know to have a non-0 gain.

If you have hardware capable of TXing with 30 dBm, this will anyway be limited by kernels-regdomain.
If for some reason you connected a negative Ant-Gain-system to the router you have to trick the OS anyway. So I'm fine to assume that the admin is aware of this setup and capable of changing the uci-setting back. A note on this possible change will be in the Releasenote, so an unwanted change can be avoided.

But if we agree, to only care for Ubiquiti-hardware, it can be refactored this way.

SvenRoederer:
@sarumpaet @booo should we just stick to NSM2 (lookup via iwinfo) or use the general approach?

SvenRoederer:
reworked it to only run on NSM2 / NSLocoM2.

passes checks on: NSM2, NSLoclM2, NSM5, WR841

sarumpaet:
:+1:
Still don't understand "On Ubiquiti it has been seen that the default TX-power on OpenWRT is at setting of Hardware, which is above regulatory limits" though, is this a typo?

SvenRoederer:
changed wording of comment

SvenRoederer:
merged in  9ac6d13

PR #89, opened 2016-10-30
Fix/issue208 (wizard autologin)

SvenRoederer:
based on @sarumpaet (https://github.com/freifunk-berlin/firmware-packages/pull/83) I just came to the idea of using a cgi.

This has the advantage of being independent of a UCI-setting
fixes https://github.com/freifunk-berlin/firmware/issues/208

### DIFF
--- a/utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
+++ b/utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
@@ -76,6 +76,22 @@ function firewall_find_zone(name)
 end
 
 
+-- checks if root-password has been set via CGI has_root-pass 
+function hasRootPass()
+	logger ("checking for root-password ...")
+
+	local isPasswordSet = true
+	local f = io.popen("wget http://localhost/cgi-bin/has_root-pass -q -O -")
+	local ret = f:read("*a")
+	if ret == "password_is_set:no" then
+		logger ("root-password not set")
+		isPasswordSet = false
+	end
+	f:close()
+	return isPasswordSet
+end
+
+
 -- Helpers --
 -- Removes a listentry, handles real and pseduo lists transparently
 function remove_list_entry(value, entry)
--- a/utils/luci-app-ffwizard-berlin/root/www/cgi-bin/has_root-pass
+++ b/utils/luci-app-ffwizard-berlin/root/www/cgi-bin/has_root-pass
@@ -1,0 +1,10 @@
+#!/bin/sh
+echo "Content-type: text/plain"
+echo ""
+USER=root
+# taken from http://serverfault.com/a/240961
+grep ${USER} /etc/shadow | awk -F":" '{ if ($2 == "!" || $2 == "") \
+		printf "password_is_set:no";\
+	else \
+		printf "password_is_set:yes"\
+	}'
--- a/utils/luci-app-ffwizard-berlin/Makefile
+++ b/utils/luci-app-ffwizard-berlin/Makefile
@@ -3,7 +3,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-ffwizard-berlin
-PKG_VERSION:=0.0.3
+PKG_VERSION:=0.0.4
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
--- a/utils/luci-app-ffwizard-berlin/Makefile
+++ b/utils/luci-app-ffwizard-berlin/Makefile
@@ -48,6 +48,8 @@ define Package/luci-app-ffwizard-berlin/install
 	$(CP) ./htdocs/* $(1)/www
 	$(INSTALL_DIR) $(1)/etc
 	$(CP) ./root/etc/* $(1)/etc
+	$(INSTALL_DIR) $(1)/www/cgi-bin
+	$(CP) ./root/www/cgi-bin/* $(1)/www/cgi-bin/
 endef
 
 $(eval $(call BuildPackage,luci-app-ffwizard-berlin))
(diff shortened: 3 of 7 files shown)


utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
```diff
@@ -76,6 +76,22 @@ function firewall_find_zone(name)
...
 
+-- checks if root-password has been set via CGI has_root-pass 
+function hasRootPass()
+	logger ("checking for root-password ...")
+
+	local isPasswordSet = true
+	local f = io.popen("wget http://localhost/cgi-bin/has_root-pass -q -O -")
+	local l = f:read("*a")
+	if l:len()==18 then
+		logger ("root-passwoird not set")
```
sarumpaet: Typo ("passwoird")

utils/luci-app-ffwizard-berlin/root/www/cgi-bin/has_root-pass
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+echo "Content-type: text/plain"
+echo ""
+USER=root
+# taken from http://serverfault.com/a/240961
+grep ${USER} /etc/shadow | awk -F":" '{ if ($2 == "!" || $2 == "*" || $2 == "") \
```
sarumpaet: This seems to return "password_is_set:no" in case root is locked, too. This seems like a bad idea to me; if someone disabled root password login, he'll have a reason for this? Check only for "".

utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
```diff
@@ -76,6 +76,22 @@ function firewall_find_zone(name)
...
 
 
+-- checks if root-password has been set via CGI has_root-pass 
+function hasRootPass()
+	logger ("checking for root-password ...")
+
+	local isPasswordSet = true
+	local f = io.popen("wget http://localhost/cgi-bin/has_root-pass -q -O -")
+	local l = f:read("*a")
+	if l:len()==18 then
```
sarumpaet: This should check the actual return string, not its length (or add comment explaining what is going on).

utils/luci-app-ffwizard-berlin/root/www/cgi-bin/has_root-pass
SvenRoederer: With a locked root-account will you be able to login to LUCI at all? 

But I can remove the "!" if you prefer.

utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
```diff
@@ -76,6 +76,24 @@ function firewall_find_zone(name)
...
 
+-- checks if root-password has been set via CGI has_root-pass 
+function hasRootPass()
+	logger ("checking for root-password ...")
+
+	local isPasswordSet = true
+	local f = io.popen("wget http://localhost/cgi-bin/has_root-pass -q -O -")
+	local l = f:read("*a")
+	-- reply can be "password_is_set:no" / "password_is_set:no"
+	-- use sting:len() to check as this is more quick than substring() ot find()
```
sarumpaet: "Performance" is the only reason? It's not like we call this function 10,000 times per second. Checking the actual string will be less brittle and more transparent.

utils/luci-app-ffwizard-berlin/root/www/cgi-bin/has_root-pass
sarumpaet: Yes please. This is how security issues happen. Also, in this case it's actually a bug: User locks root account, script says "no password set", wizard adds password to URL, webserver throws the well-known exception.

SvenRoederer:
ready for review

utils/luci-app-ffwizard-berlin/root/www/cgi-bin/has_root-pass
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+echo "Content-type: text/plain"
+echo ""
+USER=root
+# taken from http://serverfault.com/a/240961
+grep ${USER} /etc/shadow | awk -F":" '{ if ($2 == "*" || $2 == "") \
```
sarumpaet: If I understand it correctly, "*" as password means the account is locked with no previous password set ever, so we shouldn't treat this as "login possible without password" either?

SvenRoederer:
You are right on "*", but I prefer not to support empty-root passwords at all for LUCI-configpages. As this will normally open the config to everyone in the network.

sarumpaet:
"Account is locked" means "no (password based) login is possible".
In other words, if I understand correctly, if the code treats "*" as "no password set", if the user resets/reflashes, logs in via SSH, locks the root account, we'll see the well-known exception again with the current code.

SvenRoederer:
just double-cheked what the different values in shadow result in:

1. **""** (default no passowrd) - default WUI reminds of setting a password; login with empty / any password
2. **"!"** ("pass -l" called with unset password) - default WUI reminds of setting a password; login with empty / any password
3. **"pass hash"** (set via WUI / console with passwd) - default WUI shows no password-reminder; even when set to "" via passwd-tool
4. **"!pass hash"** - ("pass -l" called with set password) - default WUI no login possible

For cases 1, 2, 3 the output of the cgi reflects the "password-reminder" of pure OpenWRT. For case 4 there is no login possible at all and hard to tell about the "reminder". The cgi returns true, as something was done on the password manually, which is correct logic

With onboard-tools of our firmware there is only the "passwd"-tool to manage passwords. It only provides "lock", "unlock" and "delete" password. "lock" / "unlock" deals with the "!"

sarumpaet:
> ""!" ("passwd -l" called with unset password) - default WUI reminds of setting a password; login with empty / any password

Wow, this is quite clearly a bug in the then. `passwd -l` should disable password login regardless of whether it's done via ssh, telnet, ftp, web ui or whatever. We should probably file an upstream issue. Then I think the PR looks good.

PR #90, opened 2016-11-07
[olsrd-defaults] fix filename of olsr6 watchdog file

sarumpaet:
see
https://lists.berlin.freifunk.net/pipermail/berlin/2016-November/034556.html

We should fix this in any case (even if we use procd instead of ffwatchd) because it's a config bug anyways.


sarumpaet:
also see https://github.com/freifunk-berlin/firmware/issues/400

PR #92, opened 2016-11-08
[migration] update OWM API URL

sarumpaet:
This sets the OWM API URL to a Berlin specific one so we can have
some features like encrypted mail addresses, automated contact
forms, other map data formats, other backend databases etc.
easily without having to update all clients later.

Setting owm_api in the profile is better than changing the default
in owm.lua since the latter would (in the long run) break OWM for
other communities using Kathleen.

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
@@ -216,6 +216,13 @@ delete_system_latlon() {
   uci commit system
 }
 
+update_berlin_owm_api() {
+  if [ "$(uci get freifunk.community.name)" = "berlin" ]; then
+    log "updating Berlin OWM API URL"
+    uci set freifunk.community.owm_api="http://util.berlin.freifunk.net"
+  fi
+}
+
 fix_olsrd6_watchdog_file() {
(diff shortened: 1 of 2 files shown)


SvenRoederer:
why putting this in the assistant, which will not run by default on backbone images?
what about using https://github.com/freifunk-berlin/firmware/blob/master/patches/200-update-community-profiles.patch#L17 which we comment patch out now

sarumpaet:
Ok (see the other PR). Will merge this in a bit.

SvenRoederer:
so you plan to get this "hacky" into the 0.2.0 release and beautify for 0.2.0+, right?

utils/freifunk-berlin-migration/uci-defaults/999_freifunk-berlin-migration.sh
```diff
@@ -216,6 +216,13 @@ delete_system_latlon() {
   uci commit system
 }
 
+update_berlin_owm_api() {
```
SvenRoederer: can you also bump the version

SvenRoederer:
just found that this relates to https://github.com/freifunk-berlin/firmware/pull/404

SvenRoederer:
what's the schedule to have the owm-update function running on the new URL (http://util.berlin.freifunk.net/update_node/<node>)?

PR #96, opened 2016-11-27
owm: send release-name in field firmware.name

SvenRoederer:
currently the complete release-name is send in field firmware.release only
("firmware":{"name":"","revision":"Freifunk Berlin kathleen 0.1.2"}). With
new OpenWrt 15.05 / LuCI automatically the revision is added to this.

this makes sorting / grouping by relase difficult and requires some
extra steps on client-side.

This change corrects the datafields as follows:
"firmware":{"name":"Freifunk Berlin kathleen 0.2.0","revision":"git affe"}
- "name" is DISTRIB_ID + DISTRIB_CODENAME + DISTRIB_RELEASE
- "revision" is DISTRIB_REVISION
- use owm.get_version in place of luci.version
- rely on "/etc/openwrt_release" as original luci.version

PR #99, opened 2017-01-01
Fix deps for Freifunk-ui and openvpn-defaults

SvenRoederer:
compiles with OpenWrt & LEDE


SvenRoederer:
I plan to merge with #96

PR #102, opened 2017-03-18
add config for VPN03-via IpSec 

SvenRoederer:
IpSec works very well even it sometimes drops the connection, but is very useful to provide bandwidth of more than 10MBit/s. I've seen 47MBit/s on a EdgeRouter with nearly no systemload.

* current certificates are reused
* fits even into a 8MB-Flash router

The package will be renamed, to reflect, that this is only for VPN03-setup and 2 separate packages will be build, one for VPN03-openvpn one for VPN03-IpSec.
The wizard will still depend on VPN03-openvpn


SvenRoederer:
kathleen 0.3.0 is released. Anyone against for merging this?

SvenRoederer:
As VPN03 will be shut down "soon" this PR might help to prepare for some kind of transition to different VPNs

sarumpaet:
I can only look at this in detail next week, but from first glance, this adds a lot of mentions of "VPN03" all over the place, which isn't good for the openvpn part (which should be universal/usable with any openvpn provider in the long run), and not the ipsec part either (with "VPN03" going to disappear anyway).
Another thing, with ipsec being not very common, it would be nice to include some documentation/links in the config; surely you used some web resources when coming up with all this? Just add these as comments in the files?

SvenRoederer:
I can even split it more in something like
- ff-berlin-openvpn-default (containing network setup); to be compatible with other VPNs ant to be included by them
- ff-berlin-vpn03-openvpn (certs, openvpn-config, depending on ff-berlin-openvpn-default)
- ff-berlin-vpn03-ipsec

sarumpaet:
Logging in `60-vpn03_ipsec` should mention "ipsec".
Should we launch the IPSec thing reusing the name `VPN03` and using the Förderverein CA at all? Perhaps it'd be better to create a new FFBerlin CA and set up another instance of ca.berlin... _now_ rather than later?
Some docs would be nice, both on script level, and on package level (just a README in the root of the packages?).

SvenRoederer:
> `60-vpn03_ipsec` should mention "ipsec"
So the OpenVPN-version should also mention OpenVPN?

I'd like to see this IpSec-thing as a test  (still alpha, maybe already beta) for some handselected installations. It's not really planned for broad usage yet, as each side requires manual config on the gateway (to aviod problems with OpenVPN-instance).
It's also a bit working on https://github.com/freifunk-berlin/firmware/issues/455 to ease using other VPNs and prepare taking down of VPN03. I don't think it's the right time to setup a new VPN without knowing who will operate it and so on.

I will think of some more notes  ....

booo:
I suggest we freeze work on this until we have discussed the vpn setup in detail. Is this fine for everybody?

SvenRoederer:
The IpSec is indeed more a testbed and was created to have the setup not getting lost / personal knowledge.
That's also the reason to have the IpSec-package not in any firmware included by default.

If you are in fear we can pause here, but I see more a win in splitting the VPN03-specific stuff out of the generic VPN/Uplink-tunnel setup for the upcoming changes.

SvenRoederer:
I started to rework this VPN-mess, so lets pause this PR, but
* see https://github.com/freifunk-berlin/firmware-packages/pull/110

SvenRoederer:
I rebased it to current master.
it's based on PR #110 (split into OpenVPN and VPN03), so please do not merge before this

SvenRoederer:
As VPN03 will shutdown soon (or is expected to shutdown soon) it's not worth to merge this anymore.
I rebased PR #120 to apply on master so let's merge this and probably we will see IpSec again for tunnel-berlin ...

PR #105, opened 2017-04-11
change default hostname of unconfigured systems

SvenRoederer:
* change the default hostname of unconfigured systems

----
PR123
* do not push a default-gw to the dhcp-clients on an unconfigured router
  * this prevents multihomed-clients to have to default-gw whenn configuring the router initially (this router will probably have no uplink at this time)
  * as the "LAN"-interface config will be deleted and replaced by "DHCP" there should be no problem after later setup (ffwizzard); backbone-images need to take care if they reuse the "LAN"-interface

### DIFF
--- a/defaults/freifunk-berlin-system-defaults/Makefile
+++ b/defaults/freifunk-berlin-system-defaults/Makefile
@@ -1,0 +1,39 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=freifunk-berlin-system-defaults
+PKG_VERSION:=0.0.1
+PKG_RELEASE:=1
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/freifunk-berlin-system-defaults
+  SECTION:=freifunk-berlin
+  CATEGORY:=freifunk-berlin
+  TITLE:=Freifunk Berlin system default configuration
+  URL:=http://github.com/freifunk-berlin/packages_berlin
+  DEPENDS+= +freifunk-berlin-lib-guard
+  PKGARCH:=all
+endef
+
+define Package/freifunk-berlin-system-defaults/description
+  Freifunk Berlin configuration files for basic system-files
+endef
+
+define Build/Prepare
+	mkdir -p $(PKG_BUILD_DIR)
+endef
+
+define Build/Configure
+endef
+
+define Build/Compile
+endef
+
+define Package/freifunk-berlin-system-defaults/install
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(CP) ./uci-defaults/* $(1)/etc/uci-defaults
+endef
+
+$(eval $(call BuildPackage,freifunk-berlin-system-defaults))
--- a/defaults/freifunk-berlin-system-defaults/uci-defaults/freifunk-berlin-system-defaults
+++ b/defaults/freifunk-berlin-system-defaults/uci-defaults/freifunk-berlin-system-defaults
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+#. /lib/functions/guard.sh
+#guard "system"
+
+# change default hostname
+if [ $(uci get system.@system[0].hostname) = LEDE ]; then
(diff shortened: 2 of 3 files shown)


booo:
I think it's fine to set a default hostname. The wizzard does or should do the same. I'm not sure if "give-me-a-name" is the best default. Maybe we should derive a "unique" id and set this id as hostname?

I think it's fine to push a default route to the DHCP-client if we assume that the user connects his router to the internet via the WAN port first. I think we should first think about the setup process and review our howto and make changes according to this.

SvenRoederer:
as you mentioned there is some code in the wizard (https://github.com/freifunk-berlin/firmware-packages/blob/master/utils/luci-app-ffwizard-berlin/luasrc/controller/assistent/assistent.lua#L52). But this will not work with current LEDE-branding anymore.
Maybe we should move setting a unique default from the wizard into the "system-defaults"-package and have the wizard then changing this node-unique to a customized name.
Having this node-unique naming removed from the wizard will also make this avail for backbone-images.
What about the MAC-address as unique value? A UUID would work also, but probably requires an additional package to gernerate one.

On the DHCP-thing I can report cases, where user complained on having the Router on LAN-port breaking their WiFi-uplink on the notebook.
Think both cases are valid reasons.
Would changing the metric of the given out default route might be a solution?

booo:
What about the ULA address that is generate on first boot?

Changing the priority of the routes sounds like a plan. Only problem I see is we have to revert this change once the wizzard has run. This is kind of complicated in the backbone image.

SvenRoederer:
as noted initially, the LAN-interface will be deleted by wizard and for backbone (power-users) a note in the readme / releasnote should be fine.

SvenRoederer:
Using the ULA is also an option. Pro for MAC is, that you can look it up on the chassis.

booo:
Ok. Let's try a implementation of the metric change.

I'm not sure if we want to spread information about the MAC in the mesh. Maybe this violates privacy in some way?

SvenRoederer:
+1 for the privacy argument

SvenRoederer:
There seems to be no dhcp-option for the Route-metric (https://www.iana.org/assignments/bootp-dhcp-parameters/bootp-dhcp-parameters.xhtml)

booo:
How do we want to proceed with this?

SvenRoederer:
both ways have pros and cons ...

Let's ask the community via mailinglist ...

SvenRoederer:
I split both initial ideas into 2 separate branches to discuss them separately.
The change of the default hostname is in branch SAm0815/change_system-defaults --> PR #105 
The change of the announcement as router is in branch SAm0815/no_initial_default-route

SvenRoederer:
probably this can be used to generate a random uuid (type 4): https://gist.github.com/jrus/3197011

also uuid-type 3 or 5 are worth looking at

bobster-galore:
As a community project, I think the user should be asked for a name of her / his choice (e.g. "give-me-a-name"). 
A technical solution should be only a fallback if no name was given. As a first time user a default name like "openwrt", "lede" or something alike will be accepted because they might think it's necessary to name the router like this. So asking for a name is the most promising way also to diminish these stupid technical names like ip-address.
So may be it's a nice solution, to give as default "give-me-a-name" to be independent of lede or openwrt and go on from there.

SvenRoederer:
just adapted the ffwizard to fallback to the ip-address in case the new default hostname "give-me-a-name" is not changed by the user during setup.

SvenRoederer:
btw. the previously working code broke in LEDE-17.01 commit https://github.com/lede-project/source/commit/528b8f6f9387c9c3f10260e31a50cc53d1bd1cb4 and again changed in LEDE-17.01 commit https://github.com/lede-project/source/commit/754f474568118e124b27f80cb44a82f3a97c6ea1.

Setting the default-hostname in a "freifunk-berlin-default" script removes any dependency of the ffwizards code from the hostname-config of the underlying firmware for the future.

bobster-galore:
I think it's ok with a german version since it is not that complicated.
I suggest to use "gib-mir-einen-Namen" which is probably understandable and if not even ggl will give the correct english translation.

SvenRoederer:
on the last community-meeting I helped a new user to setup his node and it was intuitively, that he changed the name. 
So it seems time to merge ....

PR #106, opened 2017-05-30
network-defaults: disable 802.11b rates by default

SvenRoederer:
to save airtime and avoid client with bad connections
fixes https://github.com/freifunk-berlin/firmware/issues/375

### DIFF
--- a/utils/freifunk-berlin-network-defaults/Makefile
+++ b/utils/freifunk-berlin-network-defaults/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
-PKG_VERSION:=0.0.2
-PKG_RELEASE:=3
+PKG_VERSION:=0.0.3
+PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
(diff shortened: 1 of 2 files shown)


booo:
This should be part of a wifi package.

What about boards with multiple wifi interfaces?

As far as I know our current wifi configuration is part of the wizard. Maybe we should move this to the wizard code or move the code from the wizard to a default-script?

SvenRoederer:
what about adding it to the Community-profile?

SvenRoederer:
what about adding it to the Community-profile? 
There are already the Channels BSSIDs and other stuff defined.

booo:
Sounds like a plan. But I guess we still have to make a minor change to our scripts for this.

bobster-galore:
What are the necessary changes to our scripts? community profile sounds like the right place so any community can adjust to its own demand.

bobster-galore:
@SvenRoederer and me discussed the topic regarding reading community profile and let a wizard transfer the values. Since there is nothing prepared yet for a wizard and no solution visible for backbone images, we suggest to switch off 802.11b via the network-defaults-package. 
It will be documented w/ the release notes and the wiki.

SvenRoederer:
can anyone test this code on a 5GHz-router and on a dual-band?
I don't have access to any of these

SvenRoederer:
just for reference: the "legacy_rates" option will not cause any change or error on 5GHz-radios

SvenRoederer:
I close this PR in favor of https://github.com/freifunk-berlin/firmware/pull/499

PR #107, opened 2017-06-02
remove private ap from ffwizard

SvenRoederer:
this feature causes the setup to fail, even it'S disabled (issue https://github.com/freifunk-berlin/firmware/issues/427).

As no developer is willing to invest time in fixing and also the users seems not interested in this feature anymore (https://lists.berlin.freifunk.net/pipermail/berlin/2017-May/035685.html), let's simply remove it.

Code seems to work tested with https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/294


SvenRoederer:
as of the slow release-progress more users start to use teh delvelopment master, see 
* https://lists.berlin.freifunk.net/pipermail/berlin/2017-June/035963.html
* https://lists.berlin.freifunk.net/pipermail/berlin/2017-June/035896.html

I would like to push this to master, to have them not reporting known issues again and again and not getting any usefull fieldtesting and feedback.
This HotFix can then reverted by a branch with a proper fix.

booo:
Could you maybe update the last commit with the text from the pull request so people can understand the reasoning even without access to github/internet.

Apart from this it looks good!

PR #108, opened 2017-06-13
owm: disable incompatible code section

SvenRoederer:
some wifi-stuff has changed in current Lua / LuCI and breaks the whole script, that no data gets uploaded.
Disable affected section till it's fixed, to make OWM working somehow.

This is a temporary fix for https://github.com/freifunk-berlin/firmware/issues/435

### DIFF
--- a/utils/luci-app-owm/Makefile
+++ b/utils/luci-app-owm/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-owm
-PKG_RELEASE:=0.4.14
+PKG_RELEASE:=0.4.15
 
 
 PKG_BUILD_DIR := $(BUILD_DIR)/$(PKG_NAME)
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -281,6 +281,10 @@ function get()
 	local _ubusclicache = { }
 	local position = get_position()
 	local version = get_version()
+--[[
+disabled this section as of incompatible luci-code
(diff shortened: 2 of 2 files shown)


booo:
I would prefer a proper fix of the problem. Seems like a proper fix is not that difficult to implement.

bobster-galore:
quote "Seems like a proper fix is not that difficult to implement." would be nice if u could do it!
cheerio bob

booo:
@bobster-galore can you provide me with a development platform (router in the bbb-vpn)? I will try my best to come up with a solution soon.

SvenRoederer:
as of the slow release-progress more users start to use teh delvelopment master, see 
* https://lists.berlin.freifunk.net/pipermail/berlin/2017-June/035963.html
* https://lists.berlin.freifunk.net/pipermail/berlin/2017-June/035896.html

I would like to push this to master, to have them not reporting known issues again and again and not getting any usefull fieldtesting and feedback.
This HotFix can then reverted by a branch with a proper fix.

booo:
Why do they actually use the development snapshots? If it is for new fancy hardware the users should follow our advice and use tested equipment. If they don't follow this advice they should cope with the consequences.

If you as developer encourage people to test broken stuff please tell them to check out the issues before they report anything.

I don't want us to add hotfixes that we don't fix later on. Especially in this case a proper fix seems to be quite easy.

SvenRoederer:
the referenced Archer C7 and the CPe210 are supported since Kathleen 0.2.0 in General, they don't use "fancy hardware"
Sagly this is of the FCC-rules and TPLink addes some Region-lock into the Bootloader, that will be checked for factory-images.
Such support is only avail in Hedy.

SvenRoederer:
superseded by PR #109

PR #110, opened 2017-06-25
split vpn03-setup from generic OpenVPN-setup

SvenRoederer:
This splits the VPN03 specific setup from the generic setup of a tunnel via OpenVPN.

* package "openvpn-files" only has the common parts for any OpenVPN-tunnel
* new package "vpn03-files" has the VPN03-specifics (UCI-Config, ca-certificate)

This is the 1st step in phasing out the strict VPN03-binding of the firmware. In a further step separate packages for other tunnels / no tunnel can be added. see also https://github.com/freifunk-berlin/firmware/issues/455


SvenRoederer:
tested with https://buildbot.berlin.freifunk.net/builders/ar71xx-generic/builds/328 for a working VPN-setup

PR #111, opened 2017-06-25
luci-app-ffwizard-berlin: enable wifi-device by default

booo:
Enable wifi-device/radio by default in the assistent. Fixes an issue
with LEDE based builds where the 5GHz wifi-device is not enabled by
default after the wizard finished.

I'm not sure why the devices where enabled before. I can't find the
corresponding code.

Fixes https://github.com/freifunk-berlin/firmware/issues/441


booo:
Should work by now since we read the default configuration from the proper freifunk config file again.

PR #112, opened 2017-06-25
freifunk-berlin-olsrd-defaults: explicitly set IpVersion to 6 for olsrd6

booo:
If we don't set the IpVersion explicitly to 6 the startup of the daemon
fails because the IPv4 version of the txtinfo and jsoninfo plugin tries
to listen on 0::. If we don't set the IpVersion we start the IPv4
version of olsrd.

Fixes issue https://github.com/freifunk-berlin/firmware/issues/443


booo:
Maybe the IpVersion was set by older init scripts.

PR #113, opened 2017-06-26
freifunk-berlin-freifunk-defaults: use uci-default script

booo:
The freifunk-common package provides the same configuration file as the
freifunk-berlin-freifunk-defaults package. Therefore we use a
uci-defaults script to overwrite the configuration from freifunk-common
on the first boot of the router.


SvenRoederer:
what about this: https://github.com/SvenRoederer/freifunk-berlin-firmware-packages/tree/SAm0815/freifunk-defaults

booo:
As far as I remember there is a lot of junk in the configuration file **we** don't need. I would prefer a lean configuration file.

Taking the upstream file and modifying it is probably better. But for this I think we have to change the content of the upstream file first.

What about restoring the old behaviour with this pull request and fixing the overall problem later on in the process?

SvenRoederer:
to rethink for the next release is probably the best option to choose.

but the points to review still needs to be fixed / ironed out

booo:
I don't get what you mean with "still needs to be fixed/ironed out".

SvenRoederer:
> I don't get what you mean with "still needs to be fixed/ironed out".

the points to review I meintioned above

booo:
Which points?

booo:
I'm merging this to restore the old behaviour.

Please open another pull request if you want to improve on the configuration.

PR #114, opened 2017-07-09
remove http header from nc-result

andibraeu:


andrenarchy:
The headers are removed again in olsrd 0.9.6 /o\ – does the code work with both?

SvenRoederer:
with correct config / request url you can prevent these headers at all. see
* https://github.com/OLSR/olsrd/commit/93b70d54f09c730560feb48bd258d3b268f8c467
* https://github.com/OLSR/olsrd/commit/bec895abf46e4ea85c2d4b118e879cf017aaf0a6

andrenarchy:
Jaha, interesting! So what do we want? I think the HTTP headers make it more useful for other applications so I vote for enabling them and merging this PR. What's your take on this?

andibraeu:
even if we disable http headers, this PR doesn't. It makes us independent from this setting

andrenarchy:
So does this `sed` command work in both situations?

andibraeu:
unfortunately not, but I'd prefer to use the variant with headers

andrenarchy:
@andibraeu alright, can you set `httpheaders` to `true` in the [olsr setup script](https://github.com/freifunk-berlin/firmware-packages/blob/master/utils/freifunk-berlin-wizard-backend/files/usr/lib/ffwizard.d/30-olsrd.sh)? (here in this PR)

andibraeu:
so I did!

PR #117, opened 2017-08-20
wizard-backend: support arbitrary openvpn configs

andrenarchy:
As discussed [here](https://pad.freifunk.net/p/firmware-hackathon-2017-08-19) this PR adds support for arbitrary openvpn configs.


andrenarchy:
Tested with VPN03 and ipredator. The latter suffers from https://github.com/freifunk-berlin/firmware/issues/481 though but we should fix this after merging this one (it's an upstream bug).

andrenarchy:
@SvenRoederer @andibraeu any objections to merging this now?

PR #119, opened 2017-09-09
http-file-api: add package for uploading files

andrenarchy:
As outlined in https://github.com/freifunk-berlin/firmware/issues/484 we need a way to upload (large) files. This PR adds a package with a cgi script that allows file uploads with ubus session authentication+authorization (requires ubus.file.write privileges).

Example curl request:
```
curl -H 'Authorization: UBUS_SESSION' \
  -H 'Content-Type: application/octet-stream' \
  --data-binary @filename.bin \
  http://x.x.x.x/cgi-bin/files?filename=/tmp/firmware.bin
```

### DIFF
--- a/utils/http-file-api/Makefile
+++ b/utils/http-file-api/Makefile
@@ -26,7 +26,7 @@ endef
 define Build/Compile
 endef
 
-define Package/freifunk-berlin-wizard-backend/install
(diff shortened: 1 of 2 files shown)


booo:
Closing this due to inactivity.

PR #120, opened 2017-09-16
add default  package for tunnel-berlin

SvenRoederer:
* adds a new package: freifunk-berlin-uplink-tunnelberlin-files, similar to the current vpn03-package
* renames the VPN03-package to "freifunk-berlin-uplink-vpn03-files"

these new package-names need to be activated in the firmware-package-config

tested and working in then SAm0815_Hedy-alpha* builds


SvenRoederer:
see https://github.com/freifunk-berlin/firmware/commit/85df05e2a65ed44b33e4203c3937ca62ffaf852b for the changes need to be adapted on the firmware-repo

SvenRoederer:
I'd like to merge this by the end of the week (22.9.2017) if there are no objections

booo:
I would like to take a look a this and the implications first. I don't have the time this week.

SvenRoederer:
we should merge PR #122 before merging this, to have the more correct firewall-zone definition

SvenRoederer:
@booo any updates on this?
I'll merge PR #122 soon and then proceed with this PR

bobster-galore:
Seems to be nice and working, I think we should merge this.

PR #121, opened 2017-09-17
ffwizard: don't create cronjob to restart "wan"

SvenRoederer:
there seems to be no reason to do this (anymore).
This code is there since initial commit of ffwizard, but no comment for what reason

see https://github.com/freifunk-berlin/firmware/issues/344


SvenRoederer:
tested since several time in the firmware-images of branch "SAm0815_Hedy-alpha*" and no problems seen

PR #122, opened 2017-09-25
add separate firewall-zone "ffvpn"

SvenRoederer:
as the upstream-traffic of nodes with a uplink is no traffic like inside the freifunk-network, it should be considered as different from "pure freifunk".

This also supports https://github.com/freifunk-berlin/firmware/issues/416 and PR #120


SvenRoederer:
code was tested on branch "packages_PR122" of current firmware. I used the default-flavor image and tested with only "joining Freifunk-net" and also with "providing uplink".
No problems seen.

PR #123, opened 2017-10-03
don't announce as default-gw during initial setup

SvenRoederer:
By default we will not announce ourself as default-gw to the clients, to prevent them sending us packets to forward, but not being configured yet.
So the client can be connected to his fully working network (via WiFI) and connect to us for configuration (via LAN), without breaking it's current connection.
As we do this on "LAN"-interface, which will be deleted by the wizard this will not cause problems on regular installs. Only systems configured by hand and keeping the LAN-interface, have to take care.

### DIFF
--- a/defaults/freifunk-berlin-dhcp-defaults/Makefile
+++ b/defaults/freifunk-berlin-dhcp-defaults/Makefile
@@ -1,8 +1,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-dhcp-defaults
-PKG_VERSION:=0.0.1
-PKG_RELEASE:=2
+PKG_VERSION:=0.0.2
+PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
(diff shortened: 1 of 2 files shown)


SvenRoederer:
This is a discussion transferred from another issue after splitting it into smaller tasks:

@booo commented on 25 May :
I think it's fine to push a default route to the DHCP-client if we assume that the user connects his router to the internet via the WAN port first. I think we should first think about the setup process and review our howto and make changes according to this.

@SvenRoederer commented on 25 May:
On the DHCP-thing I can report cases, where user complained on having the Router on LAN-port breaking their WiFi-uplink on the notebook.
Think both cases are valid reasons.
Would changing the metric of the given out default route might be a solution?

@booo  commented on 25 May:
Changing the priority of the routes sounds like a plan. Only problem I see is we have to revert this change once the wizzard has run. This is kind of complicated in the backbone image.

@SvenRoederer commented on 25 May:
as noted initially, the LAN-interface will be deleted by wizard and for backbone (power-users) a note in the readme / releasenote should be fine.

@booo  commented on 25 May:
Ok. Let's try a implementation of the metric change.

@SvenRoederer commented on 29 May:
There seems to be no dhcp-option for the Route-metric (https://www.iana.org/assignments/bootp-dhcp-parameters/bootp-dhcp-parameters.xhtml)

@booo commented on 17 Jun:
How do we want to proceed with this?

@SvenRoederer commented on 19 Jun:
both ways have pros and cons ...
Let's ask the community via mailinglist ...

bobster-galore:
As @SvenRoederer and me discussed offline there is only a drawback at backbone images where you have to manually do what the wizard does. Elsewhere it can be considered a feature. I think we should merge it.

SvenRoederer:
on the last community-meeting I helped a new user to setup his node and he used the setup of having it'S wifi connected to the c-base lan and the router to the ethernet-port of it's notebook. I was happy to not have him to explain why he can't see the map to select the coordinates.

There was also no feedback of the community, so give it a chance and merge

PR #125, opened 2017-10-29
add ffuplink-package to use the routers-wan connection

SvenRoederer:
add a package to use the wan-connection directly (w/o VPN)

### DIFF
--- a/defaults/freifunk-berlin-network-defaults/Makefile
+++ b/defaults/freifunk-berlin-network-defaults/Makefile
@@ -2,7 +2,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
 PKG_VERSION:=0.0.4
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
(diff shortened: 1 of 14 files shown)


bobster-galore:
It would be nice to have an underhood construction which can be used by any wizard or sthg. to configure the system. Much more which kind of connection could be used.

SvenRoederer:
at the moment branch "packages/notunnel" of the firmware gets build an can then be tested.

SvenRoederer:
one issue with this code is, that the veth-link to the ffuplink interface is setup via a hotplugscript. this should be converted to a uci-setting, as we have patched the lede-17.01 netifd to support this.

SvenRoederer:
This code has been tested and verified to work with 
- OpenVPN-tunnels (at least tested by me)
- direct connection via WAN (see https://lists.berlin.freifunk.net/pipermail/berlin/2017-December/036704.html)
and I'd like to merge it soon.

Probably it's morre elegant to put the uci-setting ffberlin-uplink.uplink.auth directly into network.ffuplink.authtype.

SvenRoederer:
merged into Hedy-1.0.0-rc1

SvenRoederer:
btw: a similar feature seems to be on the way for gluon (https://github.com/freifunk-gluon/gluon/pull/1094)

PR #129, opened 2018-04-28
Fix index and shareInternet page.

nExXxuS:
due to some change in upstream projects the sites "shareInternet" and
"index" stopped working throwing an error
> A runtime error occured:
/usr/lib/lua/luci/model/uci.lua:157: attempt to call method 'get_all' (a
nil value)

Looks like we have to call get() like a method. Still i dont know why it did work before.

### DIFF
--- a/utils/luci-app-ffwizard-berlin/Makefile
+++ b/utils/luci-app-ffwizard-berlin/Makefile
@@ -3,7 +3,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-ffwizard-berlin
-PKG_VERSION:=0.0.12
+PKG_VERSION:=0.0.13
 PKG_RELEASE:=1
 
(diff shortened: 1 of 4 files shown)


SvenRoederer:
Do you remember, when it was still working?

nExXxuS:
I'm not very sure whether it still worked on 8th April build. On March 31 it worked for sure.

https://github.com/freifunk-berlin/firmware/commit/44b7806f73a46cb5cb27c1a7b2139c641f74bb65
https://github.com/freifunk-berlin/firmware/commit/df5f4f19d0ed728f88b3c78651efa917d62d74

SvenRoederer:
Probably there was a change upstream, but I will not investigate on this.
I checked the changes and they fix the page-errors also for me

SvenRoederer:
thanks

PR #130, opened 2018-05-11
Migration for v1.0.1

SvenRoederer:
This provides the migration steps to upgrade from Kathleen 0.3.0 to Hedy 1.0.0 and creates a Hedy 1.0.1.
Beside upgrading our settings, there is also a step "r1_0_0_upstream()" (https://github.com/freifunk-berlin/firmware-packages/commit/46ca9b5803df412ed595bef975a3e1c6f26e581b) which updates some files that are purely upstream and never touched by us.
With this PR we also get a UCI-value (ffberlin-uplink.preset.current) for the uplink-preset that was used last time. this can be used to determine, if the preset has been changed during the upgrade, to select the proper uplink-preset for the next time or an auto-upgrader.

### DIFF
--- a/defaults/freifunk-berlin-openvpn-files/Makefile
+++ b/defaults/freifunk-berlin-openvpn-files/Makefile
@@ -40,8 +40,6 @@ define Package/freifunk-berlin-openvpn-files/install
 	$(CP) ./openvpn/ffvpn-up.sh $(1)/lib/freifunk
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/iface
 	$(CP) ./openvpn/60-ffopenvpn $(1)/etc/hotplug.d/iface
-	$(INSTALL_DIR) $(1)/etc/config
-	$(CP) ./files/ffberlin-uplink $(1)/etc/config
 endef
 
 $(eval $(call BuildPackage,freifunk-berlin-openvpn-files))
--- a/defaults/freifunk-berlin-openvpn-files/Makefile
+++ b/defaults/freifunk-berlin-openvpn-files/Makefile
@@ -3,8 +3,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-openvpn-files
-PKG_VERSION:=0.0.7
-PKG_RELEASE:=2
+PKG_VERSION:=0.0.8
+PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
--- a/defaults/freifunk-berlin-openvpn-files/files/ffberlin-uplink
+++ b/defaults/freifunk-berlin-openvpn-files/files/ffberlin-uplink
@@ -1,4 +1,0 @@
-
-config settings uplink
-        option auth 'x509'
-
--- a/defaults/freifunk-berlin-openvpn-files/uci-defaults/freifunk-berlin-openvpn
+++ b/defaults/freifunk-berlin-openvpn-files/uci-defaults/freifunk-berlin-openvpn
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+# set set auth-type required for this uplink-type, e.g. for freifunk-wizard
+uci set ffberlin-uplink.uplink=settings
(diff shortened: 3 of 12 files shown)


pmelange:
With one upgrade with a router which also uses the bbb-vpn, the openvpn credentials (/etc/openvpn) disappeared after using this migration.  All other bbbvpn configuration (network, firewall, olsrd, openvpn) data was still there after the upgrade.

pmelange:
Unfortunately I don't have regular access to this device.  It's at a friend's house and I updated it yesterday.  Maybe, when/if  i have the time, I will set up a spare router with 0.1.2 and bbbvpn and try it out.  I'll give feedback when I do.

Have you tested with bbbvpn and 0.3.0 to 1.0.1-rc2?

SvenRoederer:
> Have you tested with bbbvpn and 0.3.0 to 1.0.1-rc2?

No, but I don't see anything in the 0.3.0 --> 1.0.1 code that is touching the cert-files

SvenRoederer:
> Have you tested with bbbvpn and 0.3.0 to 1.0.1-rc2?

No, I did not. But the only place when dealing with files in the /etc/openvpn directory is here (https://github.com/freifunk-berlin/firmware-packages/blob/migration_for_v1.0.1/utils/freifunk-berlin-migration/uci-defaults/freifunk-berlin-01-migration.sh#L307) and this is only dealing with the "freifunk_client.*" files.

sarumpaet:
@pmelange Please check `/lib/upgrade/keep.d` and `opkg list-changed-conffiles`. See https://wiki.freifunk.net/Berlin:Firmware#Wie_l.C3.A4uft_das_Firmware-Update_eines_bereits_mit_der_Berliner_Freifunk-Firmware_ausgestatteten_Routers.3F and https://wiki.openwrt.org/doc/techref/sysupgrade (bottom of page) for details.

SvenRoederer:
> Please check /lib/upgrade/keep.d and opkg list-changed-conffiles

sysupgrade -l will list all files that will be kept. 
but as far as I remember the /etc/openvpn directory is included by default (now)

bobster-galore:
I used 1.0.1 rc2 on a cpe210 which was working fine. My /etc/openvpn is still populated with bbbvpn and uplink files.

pmelange:
@bobster-galore What version of the firmware were you using before the upgrade?

@SvenRoederer I still haven't had time to set up a router with 0.1.2.  I will try to do it today, either before cbase or at cbase.  I'll let you know later.

bobster-galore:
@pmelange sorry mistook smthg, hedy  to hedy!

pmelange:
I finally got around to testing.  kathleen 0.1.2 only has freifunk_client.{crt|key} in /lib/upgrade/keep.d
Upgrading from 0.1.2 caused the bbbvpn files to dissappear.

Upgrading from 0.3.0 to 1.0.1-rcX worked as expected (bbbvpn files were still there).

SvenRoederer:
I was thinking for short in creating a 0.1.3, with a fix for this, but I can't find the OpenWrt-code we used to build 0.1.2 anymore in https://github.com/openwrt/archive

pmelange:
@SvenRoederer I don't think it is necessary to make a 0.1.3 version.  An upgrade from 0.1.2 will still loose the bbbvpn files from /etc/openvpn.  Or was there another reason why you think it makes sense to make a backport?

I agree with @bobster-galore, and that we can go ahead with the migration 1.0.1 release.

SvenRoederer:
it might have probably the same cause, as we also clean the opkg directory on the 4MB-targets.
it seems that https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=86ccaf4c18dabec9761e71a31d1456611cf1cdf9 is the fix.

For the migration, we can add these lines manually.

bobster-galore:
Todays 101rc3 worked on the wr841! It  created dnsmasq user and group. dnsmasq is up and running. Think it's ok now.

bobster-galore:
So we can release 101 now?

SvenRoederer:
I will release the code used in Hedy-1.0.0 with the added migration as v1.0.1

PR #131, opened 2018-05-22
Added the option ip4table and ip6table to the no-tunnel ffuplink interface

pmelange:
The problem is that the br-wan and ffuplink interfaces both request ip addresses via DHCP.  The second one to come up overwrites the default gw of the other.  This can lead to inconsistencies. 

This is the result of the discussion [#561](/freifunk-berlin/firmware/issues/561)

### DIFF
--- a/defaults/freifunk-berlin-openvpn-files/Makefile
+++ b/defaults/freifunk-berlin-openvpn-files/Makefile
@@ -3,8 +3,8 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-openvpn-files
-PKG_VERSION:=0.0.7
-PKG_RELEASE:=2
+PKG_VERSION:=0.0.8
+PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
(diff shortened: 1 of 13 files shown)


booo:
Can you add a comment to the code that explains the purpose of the option?

How does this change relate to the hotplug script here:

https://github.com/freifunk-berlin/firmware-packages/blob/master/uplinks/freifunk-berlin-uplink-no-tunnel/files/60-ffuplink#L12

pmelange:
The comments have been added.

About the hotplug script...  It doesn't seem to have any positive or negative effects on the system with the included lines.  If it is deemed necessary to change the hotplug script, then I am open to suggestions.

booo:
Cool!

As far as I can see the hotplug script adds a default route to the ffuplink table. The introduced ip4table option does the same thing right? I guess the part of the hotplug script that handles the defaulte route is obsolete with this change.

pmelange:
@booo @SvenRoederer I have added changes to the hotplug script, complete with comments.

SvenRoederer:
I prefer to remove all this commenting from the hotplug-script. It's so straight forward, that no comments are required to understand what's going on.

pmelange:
I removed most, but not all, of the comments.  I think it is worth keeping the few comments which are still there.

SvenRoederer:
can you clean the branch, it looks like everything is messed up

pmelange:
a new branch called ffuplink-notunnel-routing-macaddr has been created with these changes plus a static macaddr for the ffuplink_dev and migration

PR #134, opened 2018-06-10
dhcp-defaults: always announce a IPv6 default-route via us

SvenRoederer:
On Ipv6-islands, we do not have a public prefix and dnsmasq will not announce a default-route via us. This gives clients only v6-connectivity to the hosts in the local LAN-segment. 
With this option the client can also reach other hosts on the olsr-net.


booo:
I rebased and merged the commit with f4d32074dd526693470c37777faef40baa3d3fd1.

PR #135, opened 2018-07-20
Ffuplink notunnel macaddr and ipXtable

pmelange:
These two commits address issues #547 #561 and #532. 

* defines a static mac address for the ffuplink_dev
* use the network-options ip[4,6]table to set the routing table.
* drop the hotplug-script.
* migration


uplinks/freifunk-berlin-uplink-no-tunnel/uci-defaults/freifunk-berlin-notunnel
```diff
@@ -26,6 +26,13 @@ uci set network.ffuplink_dev=device
 uci set network.ffuplink_dev.type=veth
 uci set network.ffuplink_dev.name=ffuplink
 uci set network.ffuplink_dev.peer_name=ffuplink_wan
+# Create a static macaddr starting with "fe" for ffuplink devices
+local macaddr="fe"
```
SvenRoederer: gives error
> freifunk-berlin-notunnel: local: line 30: not in a function

uplinks/freifunk-berlin-uplink-no-tunnel/uci-defaults/freifunk-berlin-notunnel
```diff
@@ -26,6 +26,13 @@ uci set network.ffuplink_dev=device
...
+for byte in 2 3 4 5 6; do
+  macaddr=$macaddr`dd if=/dev/urandom bs=1 count=1 2> /dev/null | hexdump -e '1/1 ":%02x"'`
+done
+uci set network.ffuplink_dev_macaddr=$macaddr
++
```
SvenRoederer: I think there is no command "+" avail

uplinks/freifunk-berlin-uplink-no-tunnel/files/60-ffuplink
```diff
@@ -1,54 +0,0 @@
...
-    network_get_gateway gateway wan
-    ip route add default via ${gateway} dev ffuplink table ffuplink
-    logger -t ff-userlog "ffuplink connection has been established"
-  fi
-  if [ "$ACTION" = ifdown ]; then
-    logger -t ff-userlog "ffuplink connection went down"
-  fi
-fi
-
-[ "$INTERFACE" = wan ] || exit
```
SvenRoederer: Shouldn't we keep the following lines to (de)activate the ffuplink when the wan-interface changes?

pmelange: I don't believe we need it any more.

If I disconnect the cable to WAN, it waits until the DHCP lease times out in either case to loose the IP address.  But when it does, then the ffuplink dev looses it's address at the same time.

If the cable is connected and I run "ifdown wan", then the ffuplink dev immediately goes down.  When I "ifup wan", then the ffuplink dev immediately goes up.

If I change which DHCP network the wan/ffuplink is connected to, then after the lease timeout both devices obtain a new lease on the new network.

The only exception is if the connection is already up, the user changes the connection from one DHCP network to another, then runs "ifup wan"  then wan gets a new address and ffuplink will wait until the lease times out to get a new address.  This use case is also not covered in the hotplug script.

pmelange:
The commits have been squashed

uplinks/freifunk-berlin-uplink-no-tunnel/files/60-ffuplink
SvenRoederer: It'S all a bit unclear, e.g. do we want to have the uci-interface wan independen from interface ffuplink? 
Will it make sense to have the ffuplink up, when the wan is down?

Anyway, I checked on my box and dropping this script don'T seem to make it more worse ...

PR #136, opened 2018-07-24
Addon bbbdigger - an openwrt package to connect to the Berliner Backbone using tunneldigger

pmelange:
All configuration is done automatically
* opkg update; opkg install freifunk-berlin-bbbdigger

No reboot required.

* NOTE: This package depends on the ffuplink interface having "ip4table" and "ip6table" options configured


bobster-galore:
And it should be installed / configured seperately?

pmelange:
If in /etc/config/network the ffuplink interface does not have the following, then insert it. (only no-tunnel versions)
```
     option ip4table 'ffuplink'
     option ip6table 'ffuplink'
``` 

Or do you mean the package?  Once it's added to the firmware build as a module (commit 10f9d55847839523f80d4ad34726fa4ff8337278) then it will be as easy as running "opkg update, opkg install freifunk-berlin-bbbdigger'.  The postinstall script will set up all needed configuration.  IP address for meshing comes via DHCP.  The whole process is user-intervention-free.  

It could be added to webui as an optional addon.

pmelange:
I have rebased the branch with master so that all the required ffuplink (notunnel) are included.

addons/freifunk-berlin-bbbdigger/files/postinst.sh
```diff
@@ -0,0 +1,84 @@
...
+# network setup
+uci set network.${IFACE}_dev=device
+uci set network.${IFACE}_dev.macaddr=$MAC
+uci set network.${IFACE}_dev.name=$IFACE
+
+uci set network.$IFACE=interface
+uci set network.$IFACE.proto=dhcp
+uci set network.$IFACE.ifname=$IFACE
+
+uci set network.olsr_tunnel_$IFACE_ipv4=rule
```
booo: The freifunk-policy-routing script will add the ip rules. We don't have to add the rules here.

pmelange: Oh hey, that's cool.  But when is the policy routing script called?  As is it, the router does not need to be restarted when installing the package.  It just works right away.  I would really like to keep this feature.

pmelange:
@booo I made the requested changes.  Without having to reboot, the ip rules are added properly and the node quickly starts to mesh with the backbone.

Only after reboot do the extra rules for the bbbdigger interface get added to /etc/config/network.

I'm fine with the branch as it is.  It should merge without problems.  But should I first re-rebase the branch with master?  Should I squash the two commits?

Rebasing could be a problem because someone has already forked from this branch.  Same goes for squashing.  Both might leave the forked repo in limbo.

@karlbeecken would doing a rebase or squash interrupt anything you are working on?

karlbeecken:
@pmelange Do you really mean me? I don't think that I have anything to do with specially this project, so I don't know how to contribute ;)

pmelange:
@karlbeecken I mean this fork.

screenshot_2018-09-18_18-48-01

karlbeecken:
Now I'm **really** confused! I can't remember that I have forked this repo at some time, the only thing I forked was the "firmware" repo to add support for the TL-MR3020, so I never heard about this project before today.

booo:
I merged this into master so people can start beta testing this. We will refine the code over the next week e.g. add proper DNS entries for the servers and simplify the scripts a bit.

PR #137, opened 2018-07-25
uplink-notunnel: simplify creation of static macaddress

SvenRoederer:
use the functions provided by OpenWrt to format the address and set to "LAA"-bit.

* was tested in an separate script 
* looks more elegant than the loop before

### DIFF
--- a/uplinks/freifunk-berlin-uplink-no-tunnel/uci-defaults/freifunk-berlin-notunnel
+++ b/uplinks/freifunk-berlin-uplink-no-tunnel/uci-defaults/freifunk-berlin-notunnel
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+. /lib/functions/system.sh
+
 # always set correct masquerading, regardless of guard
 uci set firewall.zone_ffuplink.masq=1
 uci commit firewall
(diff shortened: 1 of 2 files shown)


uplinks/freifunk-berlin-uplink-no-tunnel/uci-defaults/freifunk-berlin-notunnel
```diff
@@ -27,11 +29,8 @@ uci set network.ffuplink_dev.type=veth
 uci set network.ffuplink_dev.name=ffuplink
 uci set network.ffuplink_dev.peer_name=ffuplink_wan
 # Create a static macaddr starting with "fe" for ffuplink devices
```
pmelange: You might want to remove/change this comment.

pmelange:
The same mac-loop is in migration.  Maybe it would make sense to add the change to migration in this PR.

pmelange:
I just saw that there is another problem in the migration script.  A small but important typo.

line 413
```
uci set network.ffuplink_dev_macaddr=$macaddr
```
should be 
```
uci set network.ffuplink_dev.macaddr=$macaddr
```

SvenRoederer:
> I just saw that there is another problem in the migration script. A small but important typo.

so just push a fix ...

PR #139, opened 2018-08-21
luci-app-owm: replace sys.net.arptable with luci.ip.neighbors in owm.lua

booo:
With a commit in the luci code base the function luci.sys.net.arptable
was replaced by luci.ip.neighbors:

https://github.com/openwrt/luci/commit/366707a681459a4d520dc97024ea0a4b3c24a326

The missing function definition caused the owm.lua program to fail.

Fixes https://github.com/freifunk-berlin/firmware/issues/507

Untested. Please test before merge.


booo:
Rebased and merge this commit on the command line.

PR #141, opened 2018-08-24
luci-app-owm: change field names for arptable in owm.lua

booo:
With a commit in the luci code base the function luci.sys.net.arptable
was replaced by luci.ip.neighbors:

https://github.com/openwrt/luci/commit/366707a681459a4d520dc97024ea0a4b3c24a326

The field names of the returned table are different to the old field
names. Therefore we have to change the field names. The values of the
table are of type userdata and we have to convert to string with the
tostring method.

### DIFF
--- a/utils/luci-app-owm/luasrc/owm.lua
+++ b/utils/luci-app-owm/luasrc/owm.lua
@@ -26,7 +26,7 @@ local table = require "table"
 local nixio = require "nixio"
 local ip = require "luci.ip"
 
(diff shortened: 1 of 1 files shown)


booo:
I tested this on a virtual machine with an olsr neighbour. The code pushed data to the owm map:

https://openwifimap.net/api/db/gib-mir-einen-namen.olsr

This should fix https://github.com/freifunk-berlin/firmware/issues/507 finally.

PR #142, opened 2018-09-16
Update Configuration for Openvpn245

pmelange:
The comp-lzo option is no longer valid.  Now the compress option must be used.  Unfortunately having something analgous to comp-lzo=no is not supported.  

The changes on the branch include the necessary uci defalt changes and the migration.  There is also a PR in firmware.

See issue freifunk-berlin/firmware#580.


SvenRoederer:
As mentioned in issue https://github.com/freifunk-berlin/firmware/issues/580 I  wonder if it will work, to have lzo enabled on the clients, but not on the servers.
Has this been tested?

pmelange:
Yes, the server and client can have different compress settings. Yes I tested it. Yes it works. There are images for anyone to do an independent test.

SvenRoederer:
I rerun a quick test with recent SAm0815_experimental build and manually applying the changes to the OpenVPN-configuration.
But I can't connect to any web-page with firefox, same on my Android-phone.

pmelange:
This is strange.  It works for me with the images from the openvpn245 branch.  Do you have any error messages?  Did you check on the server if there are any error messages?

pmelange:
A new commit has been added to openwrt-18.06
https://github.com/openwrt/openwrt/commit/d5afaa41148b363128626135bfb73b156887cddb
This may make this PR unneeded

pmelange:
This is no longer needed as of https://github.com/openwrt/openwrt/commit/d5afaa41148b363128626135bfb73b156887cddb

PR #143, opened 2018-09-27
bbbdigger: change server IP address to FQDN

pmelange:
The server has been changed from an IP address to a fully qualified domain name (a.bbb-vpn.berlin.freifunk.net).  Additionally a second FQDN has been added which currently does not resolve (b.bbb-vpn.berlin.freifunk.net).  The unresolvable domain name does not cause any performance issues but does result in a few warnings in the log.


SvenRoederer:
this PR changes 2 different things, even they relate to each other.
* changing to FQDNs
* handle multiple hostnames
it would be great to split these commits into the separate thing they do. Or change the commit message. 
This will help reviewing the code later, as the commit-message contains all facts addressed

pmelange:
The commit comment has been updated.

PR #146, opened 2018-10-21
luci-app-ffwizard-berlin: convert channel to string for uci:get

booo:
The uci:get function needs a key of type string. Therefore we convert
the channel from int to string before using it as argument for uci:get.

This fixes https://github.com/freifunk-berlin/firmware/issues/593


Akira25:
works fine. This is the section generated by the script:

config wifi-iface
        option network 'wireless0'
        option encryption 'none'
        option device 'radio0'
        option bssid 'D2:CA:FF:EE:BA:BE'
        option ssid 'intern-ch13.freifunk.net'
        option mcast_rate '6000'
        option mode 'adhoc'
        option ifname 'wlan0-adhoc-2'

PR #147, opened 2018-10-21
Revert "uplink-notunnel: simplify creation of static macaddress"

pmelange:
This reverts commit 4b379d3ea8af4fd70e62ede301ded69c92f95c4a.

The macaddress generation part of the scripts correctly flips
the U/L bit to 1, which is what we need. Unfortunately the script
does not set the LSB bit to 0.

The result is that an invalid mac address may be created, which
prevents the ffuplink and ffuplink_wan interfaces from going up.

See issue #604


SvenRoederer:
It might be more efficient to use macaddr_canonicalize()  to generate the upper 5 bytes of a macaddess than calling "dd" 5 times in a loop.
these bytes can then concat to the fixed lower byte

PR #148, opened 2018-10-24
Force tunneldigger connections to bind to the wan interface

pmelange:
If the node is not connect to wan, but meshes, tunneldigger will attept to
build the tunnel though the smart gw. This creates an undesired loopback.
The solition is to add option bind_interface 'wan'

This issue is described in https://github.com/freifunk-berlin/firmware/issues/606


pmelange:
Please don't merge this yet. I'm working on another solution which will have it's own PR-

PR #152, opened 2018-11-03
luci-app-ffwizard-berlin: set dns entry on loopback interface

booo:
Set the dns entry on the loopback interface. We set it on the loopback
interface so we only have one entry for the whole network configuration.


pmelange:
The code looks clean.  Have you tested the changes to the wizard?

So far I have put the dns server list into the loopback interface and it works fine.

pmelange:
I've tested this and it works fine.

PR #153, opened 2018-11-03
firewall-defaults: remove advanced section from firewall configuration

booo:
Some of the options are now part of the defaults section. Some are
configured in /etc/sysctl.conf.

The openwrt defaults are sane. We use these from now on.

Addresses: https://github.com/freifunk-berlin/firmware/issues/465

Maybe we should attach a migration?


pmelange:
I think we should also attach a migration.  @booo, would you like to add the migration to this PR or should we make a new one?

pmelange:
@SvenRoederer Would you take a look at this and merge it to master?

SvenRoederer:
merged in db55d052ef01e2a0f935a0341758aeb4670ed6f0 & e0302e5977e44e2d56309f2d628ef7f07972e9ed

PR #154, opened 2018-11-03
Create a tunneldigger package for freifunk-berlin

pmelange:
Create the freifunk-berlin-tunneldigger package

This package is based on the gluon tunneldigger package.
https://github.com/freifunk-gluon/packages/tree/master/net/tunneldigger

The source for the tunneldigger client is at
https://github.com/wlanslovenija/tunneldigger

The differences to the gluon version are:
* The init script is disabled by default
* Fixed problems with starting multiple tunneldigger instances
* Added a hotplug script which waits for the bind_interface to go up/down before starting/stopping any associated tunneldigger
* Does not start tunneldigger for an interface if the interface is disabled
* Does not start tunneldigger if it has already been started
* Removes pid files when tunneldigger is stopped

Additionally all changes to uplink-tunnelberlin-tunneldigger and bbbdigger have been included

The init script is disabled in the luci wizard when tunneldigger is installed in the image.  If tunneldigger is installed via opkg (for example with bbbdigger) then the init script is disabled with the postinst script

### DIFF
--- a/addons/freifunk-berlin-tunneldigger/files/tunneldigger.hotplug
+++ b/addons/freifunk-berlin-tunneldigger/files/tunneldigger.hotplug
@@ -1,0 +1,141 @@
+#!/bin/sh 
+
+. /lib/functions.sh
+. /lib/functions/network.sh
+
+PIDPATH=/var/run
+tunnel_id=0
+
+tunnel_is_up() {
+        local iface=$1
+        local pidfile="${PIDPATH}/tunneldigger.${iface}.pid"
+        local pid=0
+        local result=0
+
+        [ -e ${pidfile} ] && pid="$(cat ${pidfile})"
+        [ ${pid} != 0 ] && [ -d "/proc/${pid}" ] && result=1
+        echo "${result}"
+}
+
+interface_disabled() {
+        local iface=$1
+        local disabled=$(uci -q get network.${iface}.disabled)
+        [ -z $disabled ] && disabled=0
+        echo "${disabled}"
+}
+
+missing() {
+	logger -t td-client "Not starting tunneldigger \"$1\" - missing $2" >&2
+}
+
+handle_td_ifup() {
+	local cfg=$1
+	local enabled
+	local addresses
+	local uuid
+	local interface
+	local group
+	local limit_bw_down
+	local hook_script
+	local bind_interface
+	local broker_selection
+
+	config_get_bool enabled "$cfg" enabled 1
+	config_get addresses "$cfg" address
+	config_get uuid "$cfg" uuid
+	config_get interface "$cfg" interface
+	config_get group "$cfg" group
+	config_get limit_bw_down "$cfg" limit_bw_down
+	config_get hook_script "$cfg" hook_script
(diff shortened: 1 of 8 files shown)


SvenRoederer:
can your changes not integrated into the current tunneldigger-package from gluon? 
Does it really make sense to create another package, with a lot of duplicated code? Having our own package needs a lot of maintenance. Having a common package for gluon and freifunk-berlin in save some ressources.

pmelange:
The gluon package is just a Makefile, an example config file, and an init script.

Ours uses a similar Makefile, an example config file, and a totally different init script.  Additionally we use hotplug.

I really think it makes sense to diverge from how gluon is handling tunneldigger.

But all of that is stated above already.  I also find this easier to work with than maintaining a patch.

pmelange:
@SvenRoederer I have integrated your changes into this branch. Could you please review and merge into master?

addons/freifunk-berlin-tunneldigger/files/tunneldigger.init
```diff
@@ -0,0 +1,125 @@
...
+
+# Comment out the START declaration to prevent the init file from
+# actually starting tunneldigger at startup.  Instead rely on the
+# hotplug scripts
+#START=90
+
+PIDPATH=/var/run
+tunnel_id=0
+
+tunnel_is_up() {
```
SvenRoederer: the indentication is broken on the following lines

SvenRoederer:
@pmelange  after merging this, feel free to add cf433e82fead91bdfc8ab134e5068e4aaca63e4d to master via another PR

addons/freifunk-berlin-tunneldigger/files/tunneldigger.init
pmelange: fixed with commit edfc1fa

addons/freifunk-berlin-tunneldigger/files/tunneldigger.hotplug
```diff
@@ -0,0 +1,141 @@
...
+	config_get limit_bw_down "$cfg" limit_bw_down
+	config_get hook_script "$cfg" hook_script
+	config_get bind_interface "$cfg" bind_interface
+	config_get broker_selection "$cfg" broker_selection
+
+	let tunnel_id++
+
+	[ $enabled -eq 0 ] && return
+        [ ! -z ${bind_interface} ] && \
+                [ "$INTERFACE" != "${bind_interface}" ] && \
```
SvenRoederer: will you fix indentication here also?

SvenRoederer:
can you fixup the indentication-commits into 03cde0a8a250927ba27538125c74711469608b48 before merging?

pmelange:
The indentation in the init and hotplug scripts are fixed.

SvenRoederer:
quite huge commit. I assume it was a mixture of spaces and tabs used?

addons/freifunk-berlin-tunneldigger/files/tunneldigger.hotplug
SvenRoederer: ping

PR #156, opened 2018-11-15
ffwizard: setup 80211s mesh by default

SvenRoederer:


pmelange:
I think the value of the MeshID should be in the default file and settable in the profiles.  

Would it be possible to let the user choose between adhoc and 802.11s in the wizard?  I could image a lot of situations where a mix would be practial. eg 2.4Ghr adhoc and 5Ghz 802.11s.

I'm excited to get this into the wizard.  I keep switching to 802.11s manually.

SvenRoederer:
@bobster-galore had the idea to provide a shell-script to setup another legacy adhoc interface, if the user wants.
By default I like to see adhoc disappearing and not showing up in new nodes.

SvenRoederer:
for sure a field can be made into the profiles.
but what's the correct one. this should be discussed upstream I think. I suggest to use this code for now, an when we have defined an option in the profiles, we can add this uci_get() line.

pmelange:
@SvenRoederer I have created a new branch "feature/80211s" which does the following:

Wizard interface: 
* allows the user to choose between 802.11s and adhoc for each device
* add a descriptive name to the wireless devices ("2.4 Ghz Wifi (RADIO0)" instead of just "RADIO0")

freifunk-defaults:
* adds a new defaults section called "80211s" which can be overridden in the community profiles

Things still to do:
* modify the snippet to explain the "mesh mode" choice
* add the "80211s" section to the community profile for Berlin

I would prefer this over the hard coded version in this PR.  To include the 802.11s config defaults, I have started an issue in the luci repo.  https://github.com/openwrt/luci/issues/2503

SvenRoederer:
as mentioned above, @bobster-galore and me prefer to only have 802.11s in the wizard as some sign of a clear cut to Adhoc. Also having a migration to move from Adhoc to 802.11s.
The idea was a separate script for this switch-over. This script can also provide a simple way to setup a legacy Adhoc, if the user really needs to have it.

So we can push 802.11s forward be default in the new release and not staying on Adhoc for another decade.

This way will even not "bloat" the wizard. But feel free to have fun implementing this selection.

SvenRoederer:
what's your plan for upgraded nodes?

> I personally find the idea of having a script a bad idea. When something far simpler is required by the end user, countless emails are sent to the mailing list.

No matter from which side you look at it, you will find users who want to switch the current setup to the other mesh-type. To run the wizard will break their current setup.

Indeed this is only a side-note to think of when we give the choice of choose and not strictly bound to the issues topic.

stargieg:
Preferred mesh asap
my solution: 
https://github.com/stargieg/luci2-ffwizard/blob/master/luci2-ffwizard/files/etc/ffwizard.d/20-network.sh#L205

pmelange:
> No matter from which side you look at it, you will find users who want to switch the current setup to the other mesh-type. To run the wizard will break their current setup.

We should work on making the wizard _not_ break their current setup.  I have created issue https://github.com/freifunk-berlin/firmware/issues/661 to focus on this.

db4rne:
As long as mesh is the default I think the current implementation in branch https://github.com/freifunk-berlin/firmware-packages/tree/feature/80211s will work best for users.

pmelange:
I made PR #180 with the changes in the feature/80211s branch.

pmelange:
Can we use #180 and close this PR?

PR #158, opened 2018-11-26
Openwrt 19x: luci-master: add new dependency to luci-lib-ipkg

pmelange:
**Do not merge before the firmware is based off of OpenWRT 19.x**

This is specific for the luci-master branch

This became an issue because the luci-lib-ipkg code has been
removed from luci-base.  The upstream commit is
https://github.com/openwrt/luci/commit/4bbe32548c407ec6ae104934b3f1fe53539de9f6

### DIFF
--- a/utils/luci-app-ffwizard-berlin/Makefile
+++ b/utils/luci-app-ffwizard-berlin/Makefile
@@ -4,7 +4,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=luci-app-ffwizard-berlin
 PKG_VERSION:=0.0.16
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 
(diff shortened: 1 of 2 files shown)


pmelange:
Done, but the branch is so out of date from the master branch that both makefiles cause a conflict.  

There are currently many PR's which are modifying the makefiles.  As soon as one is committed, all the others instantly cause conflicts.  Therefore, I would prefer to wait until we move to the 19.x branch, and all other commits are merged, before attempting to resolve this conflict.

pmelange:
@SvenRoederer Can you please cherry-pick these changes into the daily/upstream-1907 branch?

SvenRoederer:
I don't see a daily/upstream-1907 branch here.

pmelange:
OK, then whatever branch is being used to build the upstream-1907 images. Perhaps you haven't tested the upstream-1907 images, but luci produces errors that these changes fix

SvenRoederer:
The firmware-barnch "daily/upstream-1907" uses master-branch of  this repo. I Don't like the  idea of adding an other branch, that needs additional maintenance ...

PR #165, opened 2018-12-09
disassoc_low_ack causes problems with clients so disable

PolynomialDivision:
* Clients can not connect
* Clients get disconnected after seconds

Related to:
https://github.com/freifunk-berlin/firmware/issues/635


pmelange:
@andrenarchy This PR modifies files in the wizard-backend.  Would you like to take a look at it?

PolynomialDivision:
Is someone still experiencing connection issues with new FR Versions? Otherwise I would jist close my PR.

pmelange:
I am sorry to say that this PR doesn't make very much sense as the "wizard-backend" was a side project which never really got going properly.  I guess it makes sense to close this PR (I'll let you click the button).

Possibly the right place to make such a change would be here:
https://github.com/freifunk/openwrt-packages/tree/master/contrib/package/community-profiles/files/etc/config

These are the profiles which are currently included in the firmware and will still be included for the foreseeable future.

SvenRoederer:
> I am sorry to say that this PR doesn't make very much sense as the "wizard-backend" was a side project which never really got going properly.

for reference: https://github.com/freifunk-berlin/firmware/issues/746

PR #170, opened 2018-12-12
uplinks: set the ffberlin-uplink.preset.previous instead of renaming …

pmelange:
Using "uci rename" does not delete the old value or what is being renamed-to.  Instead there were multiple values of ffberlin-uplink.preset.previous.  This is not the expected behaviour.  Therefore this patch was implemented.

This addresses issue https://github.com/freifunk-berlin/firmware/issues/638


SvenRoederer:
merged in https://github.com/freifunk-berlin/firmware-packages/commit/adf37cd41b2dd9c4cb8277502f787c5f6d5e5e26

PR #171, opened 2018-12-12
uplink macddr: store the macaddr of ffuplink_dev in ffberlin-uplink

pmelange:
The macaddr is saved in ffberlin-uplink.uplink.macaddr.  In this was it
will also be reused when changing uplink types back to a type which
previously had a macaddr.  This is important for those who set up special
mac filtering on their uplink routers (for example guest network or static
DHCP leases).

Situation:
install no-tunnel
change to openvpn
change back to no-tunnel

This addresses issue https://github.com/freifunk-berlin/firmware/issues/639


pmelange:
I'm fine with wait for this in 1.1.x

I agree it's a border case, but like you said, we can do it later

SvenRoederer:
The code in both uplink-packages is exactly the same. CAn you move this into a common function, that will be included by both scripts?

SvenRoederer:
merged in 48a4e4ee1c44a2f757a352fc1f481d828f1a1fbe

PR #172, opened 2018-12-16
migration: run jffs-cleanup as very last uci-default script

SvenRoederer:
move the clean up of possible duplicates of rom-files into a separate uci-default script, which will run as very last.
This prevents from problems caused by https://github.com/freifunk-berlin/firmware/issues/641, but keeps this functionality.

This was initially tested on the Hedy-1.0.x branch


SvenRoederer:
it will not fix the problem entirely, but prevents from causing problems. So it's some half of the way, till we found  a better / real solution.

SvenRoederer:
merged in master as ea6d78598a3edf4daca5c72208946364b7d1b136

PR #174, opened 2018-12-20
fix non-functioning uci:get_all() call

pmelange:
On the 18.06 branch, wireless and dhcp stopped being set up correctly.  The problem is that uci:get_all() always returns a nil table.

wizard: 
* tools.lua has a new function getMergedConfig() which handles the same functionality as needed in wireless.lua and olsr.lua.
* tools.lua has another new function mergeInto() which saves the config and either updates or creates a new section when needed.  This is taken from olsr.lua because the same code is repeated multiple times.
* wireless.lua now calls getMergedConfg() Additionally, integers have been changed to strings.
* olsr.lua now calls getMergedConfig() and mergeInto()
* generalInfo now calls getMerged() and sets each option indivudally

luci-mod-freifunk-ui
* replace calls to get_all() with individual get()s

### DIFF
--- a/utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
+++ b/utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
@@ -102,3 +102,40 @@ end
 function logger(msg)
         sys.exec("logger -t ffwizard -p 5 '"..msg.."'")
 end
+
+--Merge the options of multiple config files into a table.
+--
+--configs: an array of strings, each representing a config file.  
+--  The order is important since  the first config file is read, 
+--  then the following.  Any options in the following config files
+--  overwrite the values of any previous config files. 
+--  e.g. {"freifunk", "profile_berlin"}
+--sectionType: the section type to merge. e.g. "defaults"
+--sectionName: the section to merge. e.g. "olsrd"
+function getMergedConfig(configs, sectionType, sectionName)
+  local data = {}
+  for i, config in ipairs(configs) do
+    uci:foreach(config, sectionType,
+      function(s)
+        if s['.name'] == sectionName then
+          for key, val in pairs(s) do
+            if string.sub(key, 1, 1) ~= '.' then
+              data[key] = val
+            end
+          end
+        end
+      end)
+    end
+  return data
+end
+
+function mergeInto(config, section, options)
+  local s = uci:get_first(config, section)
+  if (section) then
+    uci:tset(config, s, options)
+  else
+    uci:section(config, section, nil, options)
+  end
+  uci:save(config)
+end
+
--- a/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/generalInfo.lua
+++ b/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/generalInfo.lua
@@ -1,6 +1,7 @@
 local uci = require "luci.model.uci".cursor()
 local sys = require "luci.sys"
 local fs = require "nixio.fs"
+local tools = require "luci.tools.freifunk.assistent.tools"
 
 f = SimpleForm("ffwizward", "", "")
 f.submit = "Next"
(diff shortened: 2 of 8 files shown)


pmelange:
This partially reverts commit e482597aad7869ae23b1b76f1f838a371480eca7

pmelange:
The problem with uci:get_all() is bigger than I though.  There are files in luci-mod-freiufnk-ui which don't work anymore and this might be the reason behind issue https://github.com/freifunk-berlin/firmware/issues/617

There are more commits to come...

utils/luci-app-ffwizard-berlin/luasrc/tools/freifunk/assistent/tools.lua
```diff
@@ -102,3 +102,22 @@ end
...
+function getMergedConfig(section)
+-- merge the values from the "freifunk" config and the community config
+  local data = {}
+  for i, config in ipairs({"freifunk", community}) do
+    uci:foreach(config, "defaults",
+      function(s)
+        if s['.name'] == section then
+          for key, val in pairs(s) do
+            if string.sub(key, 1, 1) ~= '.' then
+              data[key] = val
```
niccokunzmann: This looks a bit like duplicated code to https://github.com/freifunk-berlin/firmware-packages/pull/174/files#diff-ebf9c3e54752b0821aa3608756414bb5R97 - if there is a mistake or a future enhancement, one location can break - what do you think?
-  nothing against merge.

looks to me as one could use getMergedConfig to pass the values to uci at the other location.

pmelange: @niccokunzmann  I have force-pushed a new version of the code, without the duplicated code above.  Check it out please.

niccokunzmann: I looked at it. Duplication is removed.
If I merge some things, it is sometimes not clear, what the value is to take.

```
merge({a:1}, {a:2})
```
could be `{a:1}` or `{a:2}`.

For inheritance, the `{a:1}` would be the case, the first element having the highest priority. Yours seem to merge with the last one having the highest priority.
This should be documented as people may build on that behavior. It may however change in the future, because it is not agreed to, thus introducing bugs.

pmelange: I added this comment to the function in tools.lua. I will squash it into the commit later.  I have just started a build on the buildbot with these changes.  Let me know what you think.

```
--configs: an array of strings, each representing a config file.  The order is important since 
--  the first config file is read, then the following.  Any options in the following config files
--  overwrite the values of any previous config files. e.g. {"freifunk", "profile_berlin"}
--sectionType: the section type to merge. e.g. "defaults"
--sectionName: the section to merge. e.g. "olsrd"
```

pmelange:
I have build this locally.  It works with all uplink types.  I have created branch https://github.com/freifunk-berlin/firmware/tree/testing/18.06_master

Included in the branch are very up-to-date feeds on the 18.06 branch.  It would be nice if some people tested it.  Then maybe we can merge this into master and it will work again.

pmelange:
I have created PR https://github.com/freifunk-berlin/firmware/pull/650 which should be merged after this is merged

SvenRoederer:
> wireless.lua: ... Additionally, integers have been changed to strings.

Is the conversion to strings strictly required for this "getMergedConfig()" stuff? Would it make sense to put it into a separate commit?

SvenRoederer:
> luci-mod-freifunk-ui:  replace calls to get_all() with individual get()s

can this become a separate commit, so we can give this back to upstream?

pmelange:
>>   wireless.lua: ... Additionally, integers have been changed to strings.
>
>Is the conversion to strings strictly required for this "getMergedConfig()" stuff? Would it make sense to put it into a separate commit?

The conversion to strings guarantees compatibility will any future changes.  The original change from integer to string was in commit https://github.com/freifunk-berlin/firmware-packages/commit/e482597aad7869ae23b1b76f1f838a371480eca7.  I don't see a reason to break this into separate commits.

pmelange:
> > luci-mod-freifunk-ui: replace calls to get_all() with individual get()s
>
>can this become a separate commit, so we can give this back to upstream?

I will make a separate PR with just these changes for olsr_luci.

pmelange:
The changes to luci-mod-freifunk are in PR https://github.com/openwrt/luci/pull/2448 for the master branch and https://github.com/openwrt/luci/pull/2449 for the 18.06 branch

pmelange:
The above mentioned upstream PRs have been merged

SvenRoederer:
can you split this PR into a commit which backports the upstream stuff and another  which adds your new helpers?

This helps to follow ...

pmelange:
This has new been broken into separate commits.

SvenRoederer:
modified and merged with 36abae832685cadbf45cc8082d53be7459389c85

* beside splitting into some smaller commits, I made the spacing look like the rest of the tools.lua file
* @pmelange please don't reference to PRs only, better use commits. As commits belong to the git-repo and will also be valid when hosting on other platforms or locally.

PR #176, opened 2018-12-21
Wizard: Check for empty Mesh-IP

pmelange:
Avoids running into a lua error when leaving the "Mesh-IP" entry empty.
https://github.com/freifunk-berlin/firmware/issues/531


SvenRoederer:
added with 7edad1f32f906faa522f95ea129dce2f1bc5f5bd

PR #178, opened 2018-12-26
implement "changing uplink"

SvenRoederer:
This adds the code, which handles changing the uplink-preset.
It's done by comparing the "ffberlin-uplink.preset.current" with the definition in the currently installed preset. The UCI-value can be:
* the preset used before this image was installed
* value "undefined" when no preset was setup ever, normally a fresh install
When a real preset-change is detected the function "create_ffuplink()" is called to reinit the ffuplink-settings to factory-defaults.


SvenRoederer:
That code is quite the same as used in the "for-Hedy-1.0.x" branch.

PR #181, opened 2019-02-24
wizard: change hasRootPass to use luci.sys.user.getpass

pmelange:
The web interface has had a problem with extremely long load times and not being confronted with the changePassword page.  The problem is that the hasRootPass was using a cgi script to parse the shadow file.  Because of some change upstream, the wget call to the cgi script is timing out.

Instead, the hasRootPass function calls luci.sys.user.getpass.

The cgi-script has been removed.

This addresses issue https://github.com/freifunk-berlin/firmware/issues/665
This modifies https://github.com/freifunk-berlin/firmware-packages/commit/6c0b0643b7253a5c05dca94f8819069e68c596ed
This reverts https://github.com/freifunk-berlin/firmware-packages/commit/fe76367321b4424bb1a470e36bb2016860d38213


pmelange:
This PR is closed and PR https://github.com/freifunk-berlin/firmware-packages/pull/183 has been created to solve issue https://github.com/freifunk-berlin/firmware/issues/665 instead

PR #182, opened 2019-02-24
add pingcheck monitoring for the wan interface

pmelange:
When the wan interface looses internet connectivity, The no-tunnel uplink does not go down automatically.  Additionally, and locally generated traffic will still continue to use the wan interface (dnsmasq, statistics, ping).

uplink no-tunnel:
The scripts, when installed with pingcheck, will take down or up the ffuplink interface when the internet is not reachable via the wan interface.
    
The pingcheck package is now a dependency of freifunk-berlin-uplink-no-tunnel

network-defaults:
The scripts, when installed with pingcheck, will set routing rules for locally generated traffic depending on if the wan interface has connectivity to the internet.  When the wan interface looses connectivity, new ip rules are put in place to put the olsr-tunnel and olsr-default tables with a higher priority than the main table. When connectivity returns, the rules are removed.

The pingcheck package is now a dependency of freifunk-berlin-network-defaults

Additionally there is the migration.

This addresses issues https://github.com/freifunk-berlin/firmware/issues/659, https://github.com/freifunk-berlin/firmware/issues/660, https://github.com/freifunk-berlin/firmware/issues/623#issuecomment-439982378,

### DIFF
--- a/defaults/freifunk-berlin-network-defaults/Makefile
+++ b/defaults/freifunk-berlin-network-defaults/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
-PKG_VERSION:=0.1.0
+PKG_VERSION:=0.2.0
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
@@ -13,7 +13,7 @@ define Package/freifunk-berlin-network-defaults
   CATEGORY:=freifunk-berlin
   TITLE:=Freifunk Berlin network default configuration
   URL:=http://github.com/freifunk-berlin/packages_berlin
-  DEPENDS+= +freifunk-berlin-lib-guard +iwinfo
+  DEPENDS+= +freifunk-berlin-lib-guard +iwinfo +pingcheck
   PKGARCH:=all
 endef
 
@@ -38,6 +38,10 @@ define Package/freifunk-berlin-network-defaults/install
 	$(CP) ./hotplug.d/iface/* $(1)/etc/hotplug.d/iface/
 	$(INSTALL_DIR) $(1)/lib/functions
 	$(CP) ./lib/functions/* $(1)/lib/functions/
+	$(INSTALL_DIR) $(1)/etc/pingcheck/online.d
+	$(CP) ./pingcheck/online.d/* $(1)/etc/pingcheck/online.d/
+	$(INSTALL_DIR) $(1)/etc/pingcheck/offline.d
+	$(CP) ./pingcheck/offline.d/* $(1)/etc/pingcheck/offline.d/
 endef
 
 $(eval $(call BuildPackage,freifunk-berlin-network-defaults))
--- a/defaults/freifunk-berlin-network-defaults/pingcheck/offline.d/60-freifunk-wan
+++ b/defaults/freifunk-berlin-network-defaults/pingcheck/offline.d/60-freifunk-wan
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+[ "$INTERFACE" = wan ] || exit
+
+# Internet connectivity via WAN is down, reroute all locally generated traffic
+# over the mesh network.
+logger -t freifunk-pingcheck "WAN is down, rerouting all local traffic over the mesh network"
+ip rule add prio 3000 iif lo lookup olsr-tunnel
+ip rule add prio 3001 iif lo lookup olsr-default
+
--- a/defaults/freifunk-berlin-network-defaults/pingcheck/online.d/60-freifunk-wan
+++ b/defaults/freifunk-berlin-network-defaults/pingcheck/online.d/60-freifunk-wan
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+[ "$INTERFACE" = wan ] || exit
+
+# Internet connectivity via WAN is up, reroute all locally generated traffic
+# over WAN.
+logger -t freifunk-pingcheck "WAN connectivity is up, routing all local traffic through WAN"
+ip rule del prio 3000 iif lo lookup olsr-tunnel
+ip rule del prio 3001 iif lo lookup olsr-default
+
--- a/defaults/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-pingcheck-defaults
+++ b/defaults/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-pingcheck-defaults
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+. /lib/functions.sh
+
+# remove the sta interface, leaving only wan
+uci del pingcheck.@interface\[1\]
+
+# set the host to the anycast host livecheck.berlin.freifunk.net
+uci set pingcheck.@default\[0\].host=livecheck.berlin.freifunk.net
+
+# set the default timeout to 60 seconds
+uci set pingcheck.@default\[0\].timeout=60
+
+uci commit pingcheck
+
--- a/uplinks/freifunk-berlin-uplink-no-tunnel/Makefile
+++ b/uplinks/freifunk-berlin-uplink-no-tunnel/Makefile
@@ -2,7 +2,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-uplink-notunnel-files
-PKG_VERSION:=0.2.0
+PKG_VERSION:=0.3.0
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
@@ -14,7 +14,7 @@ define Package/$(PKG_NAME)
   CATEGORY:=freifunk-berlin
   TITLE:=Freifunk Berlin no tunnel files
   URL:=http://github.com/freifunk-berlin/packages_berlin
-  DEPENDS+=+freifunk-berlin-lib-guard +kmod-veth +freifunk-berlin-network-defaults
+  DEPENDS+=+freifunk-berlin-lib-guard +kmod-veth +freifunk-berlin-network-defaults +pingcheck
   PROVIDES:=freifunk-berlin-uplink
   PKGARCH:=all
 endef
@@ -38,6 +38,10 @@ endef
 define Package/$(PKG_NAME)/install
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(CP) ./uci-defaults/* $(1)/etc/uci-defaults
+	$(INSTALL_DIR) $(1)/etc/pingcheck/online.d
+	$(CP) ./files/etc/pingcheck/online.d/* $(1)/etc/pingcheck/online.d
+	$(INSTALL_DIR) $(1)/etc/pingcheck/offline.d
+	$(CP) ./files/etc/pingcheck/offline.d/* $(1)/etc/pingcheck/offline.d
 endef
 
 $(eval $(call BuildPackage,$(PKG_NAME)))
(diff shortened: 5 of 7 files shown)


utils/freifunk-berlin-migration/uci-defaults/freifunk-berlin-01-migration.sh
```diff
@@ -553,6 +559,7 @@ migrate () {
     r1_1_0_update_uplink_notunnel_name
     r1_1_0_remove_olsrd_garbage_collection
     r1_1_0_remove_advanced_firewall
+    r1_1_0_pingcheck
```
SvenRoederer: Do we need a migration?
Migration will only run, when the current firmware is newer than the installed config-files. as till now, there was no pingcheck package installed, there will be no configuration that needs to be migrated.

pmelange: I wasn't sure if it would set up a new package's config fine by running an upgrade or not.  

On a frest install, first the default config for pingcheck is saved in /etc/config/pingcheck.  Then at some time during the boot process the uci-defaults script is called setting up the interface and ping address to check.

By an upgrade, I'm not so sure what will happen.  
* Does the default config file from the pingcheck package get installed in /etc/config/pingcheck (a file which was not in the sysupgrade backup)?  
* Afterwards, does the uci-defaults script get called?  Notice, there isn't a guard.  Should there be?

pmelange:
It is probably not good to use the address 77.87.48.12 because of the problems with routing at the StromStr. Data Center.  [1][2][3][4][5] (there are surely more links).  Is there another universally reachable ping host that we can use?

[1] https://lists.berlin.freifunk.net/pipermail/berlin/2019-February/039220.html
[2] https://lists.berlin.freifunk.net/pipermail/berlin/2018-March/037359.html
[3] https://lists.berlin.freifunk.net/pipermail/berlin/2018-March/037450.html
[4] https://lists.berlin.freifunk.net/pipermail/berlin/2019-April/039345.html
[5] https://lists.berlin.freifunk.net/pipermail/berlin/2018-August/038173.html

pmelange:
I find it unfortunate that a PR made in Feb is now being looked at for the first time.

sarumpaet:
For dyngw we used DNS servers for ping check: https://github.com/freifunk-berlin/firmware-packages/blob/de464273780bc74928cfb8a13f27d20291b19c3b/defaults/freifunk-berlin-olsrd-defaults/uci-defaults/freifunk-berlin-olsrd-defaults#L41

pmelange:
The host 77.87.48.12 is now reachable from everywhere on the backbone (since 1 week).  So, it is not an issue (see https://github.com/freifunk-berlin/firmware-packages/pull/182#issuecomment-494028861) .  The host is ping.berlin.freifunk.net.  I would like to use this host because it means that we are relying on our infrastructure.  But certainly using one/all of the DNS servers listed in the profiles would be possible. 

One drawback from using the DNS servers is that the configuration of pingcheck would have to be added to the wizard (after the user chooses which profile to use) so each community uses their own ping host.  

Otherwise, we would have to hard code the addresses to the pingcheck configs.  This can also lead to problems like:

https://github.com/freifunk-berlin/firmware/issues/682
https://github.com/freifunk-berlin/firmware/issues/551

sarumpaet:
77.87.48.12 is just the monitor.berlin.freifunk.net host, and its ping... alias was created to have a _hostname_ not bound to the monitor VM for use in the collectd client config. It's actually likely that 77.87.48.12 will go away at some point in the future (replacing the monitor with something else, probably on a different VM, has been talked about for some while).

The idea behind using the DNS servers was to make sure to not introduce _another_ point if failure. I.e., if _all_ the DNS servers are not responding, it doesn't matter if dyngw fails as well. :) (and, at that point, we had exactly the thing happening that you're setting up now - previously, dyngw depended on the berlin.freifunk.net static IP, but that changed some years ago, with nosy's server moving)

At this point I would recommend just to use the dyngw settings and **put a comment** in there.
Just having some random IP in the config (as this PR does currently) is asking for trouble.
(that's why we added those comments to the dyngw config)

pmelange:
It seems like according to https://github.com/br101/pingcheck/blob/master/README.md only one host can be used in the pingcheck package.  Which one of the DNS servers should we use?  We could also use 8.8.8.8 or 1.1.1.1

SvenRoederer:
What about using an anycast target? this brings redundancy and low hop-count.
Back in the days I used 192.88.99.1 for 6to4-tunneling

pmelange:
8.8.8.8 is also an anycase address.  But I am fine with 192.88.99.1.  Are there any objections?

Also, I'm still not sure about the migration script.  I'm going to drop that commit from this branch.  If any problems arise, we will have to fix it later.

SvenRoederer:
then I'd prefer 8.8.8.8, as this infrastructure might be more powerful than the 6to4 gateways

utils/freifunk-berlin-migration/uci-defaults/freifunk-berlin-01-migration.sh
pmelange: migrations has been removed

pmelange:
Changes pushed

sarumpaet:
For dyngw we explicitly didn't use 8.8.8.8 because depending on Google is somewhat surprising in a project that wants to be decentralized.

Why does this have to be an IP address? Docs say 
> host: IP Address or hostname of ping destination

So why not just using `ping.berlin.freifunk.net` there?

SvenRoederer:
@sarumpaet do you think on using the DNS-entry just as some kind of alias to redirect to an IP-address and not having something hardcoded into the firmware?

Using the 8.8.8.8 or an other anycast host-set seems to me a good decentralized setup, much more than forcing our volunteer-admins to ensure 24/7 availability of one host. I remember back the shutdown of the old ping-host ...

sarumpaet:
Yes. It would be better to have multiple destinations, but if that's not supported, at least we shouldn't hardcode a Google IP.

pmelange:
How about cloudflare's 1.1.1.1 anycast address?

https://en.wikipedia.org/wiki/1.1.1.1

booo:
Please use a DNS-entry in the firmeware!!!

Maybe ping.berlin.freifunk.net should point to multiple IP addresses we control?

I don't think adding self maintained infrastructure as target for the pingcheck creates more work for us/the admins.

pmelange:
> Maybe ping.berlin.freifunk.net should point to multiple IP addresses we control?

What you propose (and correct me if i'm wrong) sound more like roundrobin dns.  I don't think that is going to help us.  If ping.berlin.freifunk.net replies with multiple addresses and the first one which is used is offline, the program using it will still use the unreachable address and produce a false negative.  

What would be the best would be our own anycast address.  Then (as far as I understand it) each machine we want to assign to ping.berlin.freifunk.net would have this anycast addr in it's configuration.

I'm not sure who can set this up for us, but if we do set up ping.berlin.freifunk.net as an anycast address, perhaps also using our partners like ffnw, then I would find this an excellent solution.

booo:
If ping.berlin.* returns multiple addresses (unicast or anycast) the addresses are returned in random order. If one of the addresses is not reachable the next ping check will test another address and at some point find an address that is reachable.

I guess we don't have the resources to announce our own anycast ip range. As far as I know you just announce the IP with BGP. But you can't announce one IP (/32). You have to announce at least a /24.

If we use a dns entry we can merge this pull request and discuss the actual IPs we use later on.

pmelange:
> If ping.berlin.* returns multiple addresses (unicast or anycast) the addresses are returned in random order.

That is true for the first time the name resolution is called, but then the results get cached and afterwards the same IP is given for each successive call.

---

Plus, and I just though of it, one of the problems we are trying to fix is the following situation.  The WAN port is connected to a DSL router which has no internet.  DNS lookups fail so using a hostname will just time out or fail and result in not having an address to ping.  For this reason, I think it may be better to have a hard coded and reliable anycast address.

sarumpaet:
The `ping` program only ever uses the first record.

> DNS lookups fail so using a hostname will just time out or fail and result in not having an address to ping.

If that really breaks the pingcheck program until reboot that's clearly a bug. Have you actually seen that behaviour?

I'm strictly against hardcoding fixed IPs in the router unless absolutely necessary, and I'm also opposed against hardcoding and relying on Google or Cloudflare IPs. I feel the latter would go strongly against the Freifunk spirit, plus it's super lame to have to tell people who are worried about privacy this: "You have to allow access to DNS and tunnel servers. And Google."

I added `livecheck.berlin.freifunk.net` (currently the monitor host; we can worry about the details later). Using a dedicated hostname (not ping.berlin) is better as ping.berlin is used for collectd latency checks and should, to keep measurements deterministic, resolve to one IP with a defined network topology, and not use anycast or anything of the like.

pmelange:
> If that really breaks the pingcheck program until reboot that's clearly a bug. Have you actually seen that behaviour?

I honestly haven't tested what would happen if a DNS lookup fails with pingcheck.  I can test it later (currently building another test) and report later.

> I'm strictly against hardcoding fixed IPs in the router unless absolutely necessary, and I'm also opposed against hardcoding and relying on Google or Cloudflare IPs

Another option would be to use the DNS root servers.  for example, i.root-servers.net -> Netnod (https://en.wikipedia.org/wiki/Root_name_server#Root_server_addresses).  They are also anycast.

> Using a dedicated hostname (not ping.berlin) is better as ping.berlin is used for collectd latency checks and should, to keep measurements deterministic, resolve to one IP with a defined network topology, and not use anycast or anything of the like.

good point about not using ping.berlin.  But relying on our own infrastructure still leaves me worried.

pmelange:
OK.  I have done some testing and here are the results.

* Using a hostname works fine
* Using a dns root server works fine (anycast address)
* The pingcheck daemon requests the dns entry every 10 pings (new roundrobin result).  It doesn't matter what the interval is set to.  I tried it with 2 second and with 10 seconds.

Assuming we will be using a hostname, here are the scenarios I can think of.

With roundrobin dns, if one destination machine is down, the pingcheck daemon could try to ping it up to 10 times.  After the 10th ping, there is a 1 in N chance of getting the same IP address as earlier (N is the number of entries in the rr-dns record).  To me, this means that the reliability of using rr-dns is dependent on __all__ the machines staying online.

With anycast, the reliability is dependant on BGP and how often the advertisements are sent out.  The default is 30 seconds for eBGP peers and 5 seconds for iBGP peers. I'm definately not an expert with BGP, but it seems to me that reliability is greater than with rr-dns.

I have set up pingcheck with an interval of 10 and a timeout of 30.  It's very improbable (not impossible), that there would be 3 consecutive false-negative pings in a row.

Tunneldigger has a timeout of 60 seconds.  Perhaps it makes sense to increase the timeout to match tunneldigger.  6 false-negative pings in a row is exponentially more unlikely than 3.

I propose using ```i.root-servers.net``` for the pinghost keeping the interval at 10 seconds and setting the timeout to 60 seconds.

---

p.s. just in case you want to know why I like i.root-servers.net is simply because I like how it reads. "I root servers" :)

pmelange:
I have force-pushed the following changes:

host=i.root-servers.net
timeout=60

@booo @sarumpaet @SvenRoederer How do you feel about this?

sarumpaet:
I changed `livecheck.berlin` to be a CNAME of that; I'd rather have berlin.freifunk.net hostnames in router configs than anything else, as we can do some live fixes if need be that way. Also, I'm not sure root-servers.net is that great as a) we don't have any SLA with them and b) nobody says those server(s) will continue to respond to ping at all, but with the livecheck thing we can postpone that discussion.

pmelange:
@sarumpaet I find that a wonderful compromise.  The changes have been pushed

pmelange:
These commits have been cherry-picked into Gonzo-1.1.0-rc1.  It would be great if these changes were merged into the master branch.

PR #183, opened 2019-03-02
uhttpd defaults: set max_requests to 3

pmelange:
Upstream, there has been a change to the default setting of uhttpd.  The
new default does not allow concurrent cgi scripts to run.  The result of this change can
be seen in issue https://github.com/freifunk-berlin/firmware/issues/665

By setting max_requests to 3 (the previous value), the hasRootPass function no longer times out and the wizard runs as expected.

The upstream commit on openwrt/master is
https://github.com/openwrt/openwrt/commit/c6aa9ff38870a30dbe6da17e4edad6039fe10ddf
The upstream commit on openwrt/18.06 is
https://github.com/openwrt/openwrt/commit/e5a0b6cde051bd24280e9c3a360e65617bce2416

This has been compiled locally and tested.

### DIFF
--- a/defaults/freifunk-berlin-uhttpd-defaults/Makefile
+++ b/defaults/freifunk-berlin-uhttpd-defaults/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-uhttpd-defaults
-PKG_VERSION:=0.0.1
+PKG_VERSION:=0.0.2
 PKG_RELEASE:=2
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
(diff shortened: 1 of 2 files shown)


pmelange:
I realize that there is currently an effort to fix the hasRootPass problem a different way (branch fix/issue665_cgi2rpc) but progress has stalled.  I recommend using this fix until another solution is working.

SvenRoederer:
PR https://github.com/freifunk-berlin/firmware-packages/pull/186 it has been confirmed to work. Also upstream has reverted the change leading to the problem fixed here (f00a4ae6e0b1f7b3d84e11e8dc4dd562088584e0, e0505cc0182acf569cdd7b849189c134684160f8).

So I suggest to close this unmerged.

PR #185, opened 2019-03-17
disable ffuplink-interface during initial setup

SvenRoederer:
On a fresh installed system, there is no valid setup of ffuplink. So we disable this interface, till it's configured by some uplink-preset.
The advantage of this change s, that an unconfigured ffuplink (as so uplink-preset was installed for manual configuration) will not cause any unintended sideeffects.

This code ha been tested on the SAm0815_experimental branches since some time.

### DIFF
--- a/defaults/freifunk-berlin-network-defaults/Makefile
+++ b/defaults/freifunk-berlin-network-defaults/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
-PKG_VERSION:=0.1.0
+PKG_VERSION:=0.1.1
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
--- a/defaults/freifunk-berlin-network-defaults/Makefile
+++ b/defaults/freifunk-berlin-network-defaults/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
-PKG_VERSION:=0.1.1
+PKG_VERSION:=0.1.0
 PKG_RELEASE:=1
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
--- a/defaults/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-ffuplink-defaults
+++ b/defaults/freifunk-berlin-network-defaults/uci-defaults/freifunk-berlin-ffuplink-defaults
@@ -15,3 +15,5 @@ uci set ffberlin-uplink.uplink.auth=none
 uci commit ffberlin-uplink
 
 create_ffuplink
+uci set network.ffuplink.disabled=1
+uci commit network.ffuplink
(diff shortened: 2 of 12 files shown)


pmelange:
I just did a test.  I installed the tunneldigger version.  In the wizard, I selected mesh only (sharenet=0).  I also connect the wan port to the internet.  After reboot, the tunneldigger interface builds.  That means that disabled=0 for the ffuplink interface.

I set up a second node with the no-tunnel version.  In the wizard, I again selected mesh only.  I also connected the wan port to ethernet.  After reboot, the dhcp interface builds.

Both nodes see each other as smart gateways.  There is no internet.

Possibly, instead of setting disabled=0 in the uci-defaults, it shold go in the wizard.  Maybe where sharenet is set to 0.

pmelange:
An alternative to this PR has been created (#190)

pmelange:
@SvenRoederer can we close this PR and use #190 instead?

SvenRoederer:
@pmelange  essentially the difference between this PR and PR #190 is the place where the interface will be enabled.
This PR leaves it to the uplink-scripts:
* advantage: installing a uplink-package on the manual / backbone image will enable the ffuplink-interface automaticaly

PR #190 relies on the wizard:
* advantage: the ffuplink is inactive as long as "share internet" in not active

It's hard to decide for one as the best solution ...

pmelange:
> @pmelange essentially the difference between this PR and PR #190 is the place where the interface will be enabled.
> This PR leaves it to the uplink-scripts:
> 
>     * advantage: installing a uplink-package on the manual / backbone image will enable the ffuplink-interface automaticaly
> 
> 
> PR #190 relies on the wizard:
> 
>     * advantage: the ffuplink is inactive as long as "share internet" in not active
> 
> 
> It's hard to decide for one as the best solution ...

Try the tests described in https://github.com/freifunk-berlin/firmware-packages/pull/185#issuecomment-474150669.  

After the wizard is ran and "share internet" is not enabled, then the node should not announce a smart gateway.  PR #185 Fails this test.  PR #190 does not fail this test.

SvenRoederer:
> I just did a test. I installed the tunneldigger version. In the wizard, I selected mesh only (sharenet=0). I also connect the wan port to the internet. After reboot, the tunneldigger interface builds. That means that disabled=0 for the ffuplink interface.
> 
> I set up a second node with the no-tunnel version. In the wizard, I again selected mesh only. I also connected the wan port to ethernet. After reboot, the dhcp interface builds.

From the logic perspective it should be:
* the user installs an uplink-package
  * the user wants to use an uplink and is willing to share it. what would be the reason for installing an uplink-package anyway.
  * this installed uplink should work out of the box, when possible. This will work for tunneldigger and notunnel but others might need additional configuration (e.g. certificates for OpenVPN)
* When no uplink-package is installed, the use will for sure not share
  * This does not relate to the status of the WAN-connetction of the router itself. As we separated WAN and ffuplink, the router can use the WAN-connection for its traffic (statistics, updates, DNS) but all client-traffic should go only via mesh

By writing this, I feel we should rework this system, as soon as we implement the "common Image for dynamical selection of uplink-type".

@pmelange I suggest ti use this PR, as it now combines your suggestion with enabling the ffuplink-interface in the wizard and also keeps the idea to have the uplink-packages enable the uplink. This way we can later revert individual changes, as we need them.

pmelange:
Quite honestly, do whatever you want with this.  I am no longer developing for the freifunk firmware.

everloop2 (approved):

PR #189, opened 2019-06-14
migration: remove cleaning non-profile config files from jffs system

pmelange:
The removal of non-profile config files breaks the ability to configure any remaining and as yet untouched configurations.  The result of trying to modify one of the config files removed in the /overlay filesystem results in ```uci: I/O error```

This has been documented in https://github.com/freifunk-berlin/firmware/issues/641

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/zz_freifunk-berlin-migration-clean_jffs.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/zz_freifunk-berlin-migration-clean_jffs.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
 
-# delete any overlay config files duplicated from romfs by sysupgrade - saves JFFS2 space
-cd /overlay/upper/etc/config/ && for i in *; do [ -f "$i" ] && if cmp -s "$i" "/rom/etc/config/$i"; then rm -f "$i"; fi; done;
+# delete any overlay profile config files duplicated from romfs by 
+# sysupgrade - saves JFFS2 space
+cd /overlay/upper/etc/config/ && for i in profile_*; do [ -f "$i" ] && if cmp -s "$i" "/rom/etc/config/$i"; then rm -f "$i"; fi; done;


pmelange:
These commits have been cherry-picked into Gonzo-1.1.0-rc1. It would be great if these changes were merged into the master branch.

SvenRoederer:
There is an upstream change to add a ```-u``` to sysupgrade (https://git.openwrt.org/?p=openwrt/openwrt.git;a=commit;h=20b23270b712ba0c22e204b1347f04b2206d7018) which seems to handle this issue too, probably we can use this option as default and drop our code completely.

pmelange:
We would just have to patch https://github.com/openwrt/luci/blob/ab70fb1ac6cd48539285dfaae7d28d29f487d8c3/modules/luci-mod-system/luasrc/controller/admin/system.lua#L191

from 
```
	local keep = (http.formvalue("keep") == "1") and "" or "-n"
```
to
```
	local keep = (http.formvalue("keep") == "1") and "-u" or "-n"
```

It is a simple small patch.  I think it is worth considering.

People using the command line would have to know to use the -u option.

SvenRoederer:
What about changing the default of sysupgrade?

pmelange:
Actually, I don't think we should delete the files in any case or to patch sysupgrade either.  Since dropping 4MB is a fact now, I don't see why we should be trying to save this little amount of space on the jffs partition at all.

SvenRoederer:
As it's a bug, to have the profiles in the backup. Following situation:
Node created with Hedy-1.0.0. The profiles get updated and a new release gets published. The node gets updated, the UCI-files (incl. profiles) are saved and restored during upgrade. So as result, the recent profiles are in /rom only and the old files from the backup are used in place.
This also breaks the login @booo did with updating the DNS-Servers on every release.

SvenRoederer:
btw. this is a backport for OpwnWrt18.06 of the -u switch https://github.com/SvenRoederer/freifunk-berlin-firmware/commit/21d888896364110891d6d071b57f405ea105958a

pmelange:
Then the obvious solution is to copy all the profiles from rom to overlay in the migration script.  Also in the migration script, the freifunk config file should be handled specially as it is a mixture of default settings and running system settings.

SvenRoederer:
migration will not apply to the backbone-images, whereas the upstream-solution of sysupgrade will allso work there.

pmelange:
Not with /etc/config/freifunk.  The contact and community is set, so it has been updated.  But if there is a new release with new defaults in /rom/etc/config/freifunk, then only the migration script will be able to import it.

here is the pseudocode
save uci show freifunk.conact
save uci show freifunk.community
cp /rom/etc/config/freifunk /etc/config/freifunk
write saved values

pmelange:
Here it is

```
migrate () {
  log "Migrating from ${OLD_VERSION} to ${VERSION}."

  log "Updating Community-Profiles."
  cp /rom/etc/config/profile_* /etc/config

  log "Importing updates to /etc/config/freifunk"
  CONTACT=$(uci show freifunk.contact)
  COMMUNITY=$(uci show freifunk.community)
  cp /rom/etc/config/freifunk /etc/config
  OLDIFS=$IFS
  IFS=$'\n'
  for i in $CONTACT ; do
    $(uci set $i)
  done
  for i in $COMMUNITY ; do
    $(uci set $i)
  done
  IFS=$OLDIFS

  if semverLT ${OLD_VERSION} "0.1.0"; then
...
```

pmelange:
commit https://github.com/freifunk-berlin/firmware-packages/commit/afcc178f9ae4dbc906157eea432bd83a1fc34476 is the new proposed solution.  As this PR does not address the real problem, I am closing it.

pmelange:
update https://github.com/freifunk-berlin/firmware-packages/commit/ca2727037cacad43892ba2f00818c71fef00c70d

PR #190, opened 2019-06-14
disable ffuplink-interface during initial setup - revisited

pmelange:
This is an alternative to PR #185 

In this version, the option network.ffuplink.disabled is set to 0 in the wizard if sharenet is selected.  Additionally, if sharenet is disabled (for example re-running the wizard), the same option is set to 1.

Quoted from https://github.com/freifunk-berlin/firmware-packages/pull/185#issue-261869574
> On a fresh installed system, there is no valid setup of ffuplink. So we disable this interface, till it's configured by some uplink-preset.
> The advantage of this change s, that an unconfigured ffuplink (as so uplink-preset was installed for manual configuration) will not cause any unintended sideeffects.

This implementation has been tested and passes the tests described in https://github.com/freifunk-berlin/firmware-packages/pull/185#issuecomment-474150669

### DIFF
--- a/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/wireless.lua
+++ b/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/wireless.lua
@@ -88,6 +88,11 @@ function main.write(self, section, value)
     --share internet was not enabled before, set to false now
     uci:set("ffwizard", "settings", "sharenet", "0")
     uci:save("ffwizard")
+    --in case the wizard has been re-ran, ensure ffuplink is disabled
+    uci:set("network", "ffuplink", "disabled", "1")
+  else
+    --sharenet was enabled, therefore enable the ffuplink network interface
+    uci:set("network", "ffuplink", "disabled", "0")
   end
 
   -- store wizard data to fill fields if wizard is rerun
--- a/defaults/freifunk-berlin-network-defaults/Makefile
+++ b/defaults/freifunk-berlin-network-defaults/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-network-defaults
-PKG_VERSION:=0.1.0
+PKG_VERSION:=0.1.1
 PKG_RELEASE:=1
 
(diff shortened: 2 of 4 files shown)


utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/wireless.lua
```diff
@@ -88,6 +88,11 @@ function main.write(self, section, value)
     --share internet was not enabled before, set to false now
     uci:set("ffwizard", "settings", "sharenet", "0")
     uci:save("ffwizard")
+    --in case the wizard has been re-ran, ensure ffuplink is disabled
```
SvenRoederer: why are these changes made in the wireless.lua file? As it acts on the "share internet"-settings it should be better placed in "shareInternet.lua".

pmelange: shareInternet.lua is __only__ called when the user clicks on " Am Freifunk-Netz teilnehmen und Internet teilen" https://github.com/freifunk-berlin/firmware-packages/blob/8e2b55e1e4e9b396562cb0305cdecd912a5f7cda/utils/luci-app-ffwizard-berlin/luasrc/view/freifunk/assistent/decide.htm#L37  

If the user clicks "Am Freifunk-Netz teilnehmen" https://github.com/freifunk-berlin/firmware-packages/blob/8e2b55e1e4e9b396562cb0305cdecd912a5f7cda/utils/luci-app-ffwizard-berlin/luasrc/view/freifunk/assistent/decide.htm#L19 then shareInternet.lua is jumped-over and the wizard goes straight to "optionalConfigs".

When the wizard is run for a second time, and if this code were in "shareInternet", and the user changes from sharing to not sharing, then the code would never run and ffuplink would not be disabled.

Sure the code could be moved to https://github.com/freifunk-berlin/firmware-packages/blob/8e2b55e1e4e9b396562cb0305cdecd912a5f7cda/utils/luci-app-ffwizard-berlin/luasrc/model/cbi/freifunk/assistent/optionalConfigs.lua#L10-L13 with an else clause, but that still is not in "shareInternet".  Therefore "shareInternet.lua" is not a better place.

PR #194, opened 2019-12-27
rework package "Uplink tunnel-berlin"

SvenRoederer:
* combine both variants (OpenVPN, Tunneldigger) in a single package
  * there is only one folder per uplink
* rename the current packages to have the same "base-name" and a variant-specific suffix


SvenRoederer:
after merging this, the firmware-repo need to get added https://github.com/freifunk-berlin/firmware/commit/a31680bea14e7fb5bfcba66a3506ba103fa392d5

SvenRoederer:
merged in 3e2bd4d8ed82879ad3ba3c5c19c996b5d2516ba0, 7e33a8783e2d21b710b43e22a141ee774c9b290

PR #196, opened 2020-02-29
collectd-addons/snmp: scale uptime to days

SvenRoederer:
* sysUpTimeInstance is measured in 100/second
* scale it down to more meaningful days of uptime (as it's done for collectd-uptime by default)

### DIFF
--- a/addons/collectd-addons/files/collectd-snmp/ubnt-airmax.conf
+++ b/addons/collectd-addons/files/collectd-snmp/ubnt-airmax.conf
@@ -5,7 +5,8 @@ LoadPlugin snmp
     <Data "ifmib_uptime">
 	Type "uptime"
 	Table false
-	Instance "system"
+	Instance "days"
+	Scale 1.15740740741e-07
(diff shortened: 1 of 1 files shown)


SvenRoederer:
checkout at: http://monitor.berlin.freifunk.net/detail.php?p=snmp&t=uptime&h=Verklaerung-core&s=604800&x=800&y=350

pmelange:
Shouldn't there be a migration for this?  Otherwise, I find the idea good.

SvenRoederer:
As the date with the new scaling gets added to the existing chart the old data "crash" the visualization. See picture below:
* orange graph is unscaled data
* red graph is scaling to seconds
* blue graph is scaling to days
You hardly can see the graph for seconds-scale, but the days-scale is like "0".
Screenshot_20200407_224335

So I suggest not to change an existing setup automatically.

SvenRoederer:
#214 also updates this template and should merge timely close.

everloop2 (approved):

PR #197, opened 2020-04-01
change hardcoded ip to actual dns name

rtznprmpftl:


SvenRoederer:
The migration for this is still missing

pmelange:
> The migration for this is still missing

Please see #198

PR #198, opened 2020-04-02
migration: add hostname for statistics server

pmelange:
This is a follow-up to fd19100f5fa56635889cb78109dbc4a555f79d99

### DIFF
--- a/utils/freifunk-berlin-migration/uci-defaults/freifunk-berlin-01-migration.sh
+++ b/utils/freifunk-berlin-migration/uci-defaults/freifunk-berlin-01-migration.sh
@@ -491,6 +491,12 @@ r1_1_0_firewall_remove_advanced() {
   config_foreach firewall_remove_advanced advanced
 }
 
+all_monitor_server() {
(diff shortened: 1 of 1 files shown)


SvenRoederer:
I wonder why this adds a new unconditional section for this migration-setting. Why not adding into a release-specific section, as we always did before?
Is there an advantage?
This will change the current logic and might cause confusion.

pmelange:
first of all, all images, even if they are not a "release version" will perform this migration.  Second, this and possibly other changes, should not necessarily depend on version number.  The first thing that comes to mind are DNS servers.

SvenRoederer:
> Second, this and possibly other changes, should not necessarily depend on version number.

have in mind, that when adding something after a release, this older releases will not benefit from this migration. (adding this for the upcoming release "1.1.x" / "2020.04.x" will run this migration when it's in the correct section. This migration will also run when in the generic section. If you upgrade from a old release to the previous stable, there is still the old migration-script, which don't knows of the changes for the upcoming release. So it will not be migrated on older releases, no matter if in a release-section or in a generic-section, anyway. As the code is not there.)

PR #199, opened 2020-04-05
bbbdigger: use uci-default script in place of separate shell script

SvenRoederer:
Since some time the uci-default scripts are ran after installing a package with opkg. So we can use this method also to trigger the runtime config-setup and can drop doing it with the PostInstall-script.

This will make the code more "clean".


pmelange:
According to https://openwrt.org/docs/guide-developer/packages, a post-install script is what should be used.  After looking at the various packages in openwrt/packages, post-install scripts are used extensively.  I have yet to find a package which makes use of uci-defaults exclusively.  What I have seen, are packages which put the contents of the post-install script in a uci-default in case a package is burnt into an image.

This package should not be burnt into the freifunk images.  It is an add-on.

I think that we should stay with postinst.

Regardless, this PR needs to be changed to add an ```exit 0``` at the end of the uci-default script.

SvenRoederer:
just updated the OpenWrt-wiki regarding the uci-defaults. The feature was added in https://github.com/openwrt/openwrt/commit/9b9c78e071469313566473984fb79f54c1881c10

everloop2 (approved):

PR #201, opened 2020-04-09
bbbdigger/postinstall: suppress output of interface-identifier

Akira25:
This commit will suppress the output of the interface-identifier at installation of package bbb-digger.

This should solve Issue freifunk-berlin/firmware#675

### DIFF
--- a/addons/freifunk-berlin-bbbdigger/Makefile
+++ b/addons/freifunk-berlin-bbbdigger/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=freifunk-berlin-bbbdigger
-PKG_VERSION:=0.0.3
+PKG_VERSION:=0.0.4
 PKG_RELEASE:=0
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
--- a/addons/freifunk-berlin-bbbdigger/files/postinst.sh
+++ b/addons/freifunk-berlin-bbbdigger/files/postinst.sh
@@ -14,6 +14,11 @@ TUNNEL_SERV='a.bbb-vpn.berlin.freifunk.net:8942 b.bbb-vpn.berlin.freifunk.net:89
 IFACE=bbbdigger
 BIND=wan
 
+#CONF for logging
+LOG_NAME='bbbdigger_install:'
+LOG_LEVEL=3
+
+
 # tunneldigger UUID (and MAC) generation, if there isn't one already
 # See the website https://www.itwissen.info/MAC-Adresse-MAC-address.html
 MAC=$(uci -q get network.${IFACE}_dev.macaddr)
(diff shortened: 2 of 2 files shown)


pmelange:
Hi @Akira25 ,

It would be good to bump the version number in https://github.com/freifunk-berlin/firmware-packages/blob/master/addons/freifunk-berlin-bbbdigger/Makefile

Additionally, can you post the console output from installing the package?

Thanks for the good work

SvenRoederer:
What about the concerns mentioned in https://github.com/freifunk-berlin/firmware/issues/675#issuecomment-478701919?

Akira25:
@SvenRoederer I understood that @pmelange doesn't want to pipe the output of the whole script into `/dev/null`. I agree, that the addition of an interface via uci could fail. But I think it is very unlikely to happen. 
If so, we should get back on this issue.
```
root@172-16-6-1:/tmp# opkg install freifunk-berlin-bbbdigger_0.0.4-0_all.ipk 
Installing freifunk-berlin-bbbdigger (0.0.4-0) to root...
Configuring freifunk-berlin-bbbdigger.
Stopping tunneldigger for interface ffuplink
  tunneldigger stopped
Starting tunneldigger "ffuplink" on ffuplink
Starting tunneldigger "bbbdigger" on bbbdigger
root@172-16-6-1:/tmp#
```

SvenRoederer:
issue https://github.com/freifunk-berlin/firmware/issues/675 is clearly about :

>I wonder on the "cfg164d76" thing

Obviously it doesn't make sense to try to suppress the opkg-message of installation-status

Akira25:
@pmelange do you think we should add a hook which checks for the proper recreation of that interface? I'm not sure if that wouldn't be a bit overdriven, as the prior version did not prompted any meaningful information on that.

pmelange:
Maybe the best option would be to log all of the output of the script to the logging subsystem.

In the extremely rare case that there is a failure, the script could tell the user to look at the log.

SvenRoederer:
@pmelange as the uci-staging area is in /tmp (tmpfs) the most likely error might be no RAM. When the system runs out of RAM:
* a runtime failure of script-setup will be the most minor problem 
* adding to the RAM-based logbuffer will also fail very likely

I also had a short look for "uci add" in the packages feed and have seen nowhere a check like you suggest.

pmelange:
The only other possible source for a failure would be that the overlay filesystem is full.  It's probably a  lot more likely that running out of RAM.  And in the case where the overlay filesystem is full, then the admin on the router has a lot more to deal with than the bbbdigger connection to the backbone not work.

I still believe that failure is an extreme edge case.  The fact that the "uci add" call prints out this annoying cfg164d76 so far has not caused (I think) anyone writing to the mailing list wonder what the %*?# is going on.

simply by changing the line
```
uci add olsrd Interface
```
to 
```
RES=$(uci add olsrd interface)
```
removes the annoying text.  But it doesn't handle any failures (which is done by checking the value of $?).

Akira25:
The current commit suppresses the output of the id and prompts in case of failure an informative line in the log as on std as well. If we agree on, that the case of failure is too unlikely, we could also drop the if-statement and the logger-line.

SvenRoederer:
> The only other possible source for a failure would be that the overlay filesystem is full. It's probably a lot more likely that running out of RAM. 

As mentioned the uci changes stay in the tmpfs at first and don't cause any write-access to flash. The changes will be written to flash with ```uci commit```.

PR #202, opened 2020-04-10
owm.lua: add help-text.

Akira25:
This PR gives a proposal for a solution of Issue Freifunk-berlin/firmware#485.

### DIFF
--- a/utils/luci-app-owm/files/owm.lua
+++ b/utils/luci-app-owm/files/owm.lua
@@ -6,6 +6,32 @@ require("luci.sys")
 require("nixio.fs")
 require("luci.httpclient")
 
+-- Print help text
+function print_help()
+	print("owm.lua - Tool for registering routers at openwifimap.net\n")
+	print("Options:\
+        -\-help|-h:\tprint this text\
+	\
+        -\-dry-run:\tcheck if owm.lua is working (does not paste any data).\
+	\t\tWith this option you can check for errors in your\
+	\t\tconfiguration and test the transmission of data to\
+	\t\tthe map.")
+	print("\nIf invoked without any options, this tool\
+will try to register your node at the community-map,\
+but will execute silently. To work correctly, this tool\
+will need at least the geo-location of the node (check\
+with -\-dry-run).\
+\
+To override the server used by this script, set freifunk.community.owm_api.\
+")
+end
+
+-- check for arguments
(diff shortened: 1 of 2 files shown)


pmelange:
The server name can be overridden in the freifunk config file. Maybe that should be mentioned in the help text. Aside from that, there are options that are required to be able to work. They also might be worth mentioning

Otherwise, I think the changes are cool

Akira25:
Thank you for the advice. I added those information in the text.

utils/luci-app-owm/files/owm.lua
```diff
@@ -6,6 +6,34 @@ require("luci.sys")
...
+	\
+        -\-dry-run:\tcheck if owm.lua is working (does not paste any data).\
+	\t\tWith this option you can check for errors in your\
+	\t\tconfiguration and test the transmission of data to\
+	\t\tthe map.")
+	print("\nIf invoked without any options, this tool will try to register your node at\
+the community-map, but will execute silently. To work correctly, this tool will need\
+at least the geo-location of the node (check with -\-dry-run).\
+\
+To send the data to another server, you can give its URL in /etc/config/freifunk (option: owm_api).\
```
pmelange, line 23: Maybe change the text to "To override the server used by this script, set freifunk.community.owm_api" or something similar.  That way, the config filename and the section are clear and can also be easily copied into a call to "uci show" or "uci set".

pmelange:
Don't forget to bump the version number in the Makefile

Otherwise, it's looking better and better.

utils/luci-app-owm/files/owm.lua
Akira25: thank you. That's my aim. :) I changed the sentence as you requested.

pmelange: Yeah!

pmelange:
I just ran these changes on an already running system.  The indentation is not working very well on a 80 char terminal.  I suggest putting the newline after "this tool"

Additionally, I accidentally ran it with the option --hellp.  Well, the script ran like normal.  Perhaps (but it's not important) it would make sense to check for invalid options (causing -h and --help to also fall into this category).

pseudocode:
```
if (args > 0 && arg[1] !0 "--dry-run")
   print help
   exit
fi
```

Akira25:
The text should now work also an a 80-char terminal. I added the feature to secure against wrong options.

PR #203, opened 2020-05-01
define an internal "Non release flag" 

SvenRoederer:
Define a uci "system.was_development" flag, which signals, that the node has ran development code once. This flag might be useful to signal that the system might have conflicting settings, due to an specific development-branch. 
A development-branch might define a setting, which is different in the final release, but can potentially cause malfunction. As of our long release-pauses also regular users have to use development-firmware, to support their hardware (a.t.m. >100 such nodes online). 
Very likely many of these users run such firmware, as it's their only option. When a release will be ready, these people very likely just upgrade to this release. I don't know if there is a specific version around where this will cause problems, now or even in a further release.

As a simple example I made a check into the migration-script to not run automatically on such nodes. The migration can be started manually and checked for running correct.

### DIFF
--- a/defaults/freifunk-berlin-system-defaults/uci-defaults/freifunk-berlin-system-defaults
+++ b/defaults/freifunk-berlin-system-defaults/uci-defaults/freifunk-berlin-system-defaults
@@ -1,12 +1,22 @@
 #!/bin/sh
 
 . /lib/functions/guard.sh
+. /etc/openwrt_release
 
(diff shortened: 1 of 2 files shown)


everloop2 (approved):

SvenRoederer:
Thanks for merging. based on this I'll look for the code that made the wizard not automatically reboot on developer-systems (for easy review of changes.

PR #204, opened 2020-05-06
add new package "freifunk-berlin-lua-tools"

SvenRoederer:
Move some generic code from the assistent into this new common package.
This way the functions can be used by other code as well.

### DIFF
--- a/addons/freifunk-berlin-lua-tools/Makefile
+++ b/addons/freifunk-berlin-lua-tools/Makefile
@@ -1,0 +1,16 @@
+
+include $(TOPDIR)/rules.mk
+
+PKG_RELEASE:=1
+
+LUCI_TITLE:=Freifunk Berlin lua tools
+LUCI_DESCRIPTION:=common helper tools used by Freifunk-Berlin packages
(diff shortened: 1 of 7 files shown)


SvenRoederer:
I have not checked if the wizard still works after this change, but feedback on the overall idea is very welcome.

addons/freifunk-berlin-lua-tools/Makefile
```diff
@@ -0,0 +1,13 @@
+
+include $(TOPDIR)/rules.mk
+
+LUCI_TITLE:=Freifunk Berlin lua tools
+LUCI_DEPENDS:=luci-lib-base
```
PolynomialDivision, line 5: +luci-lib-base ?

SvenRoederer: > +luci-lib-base ?

https://github.com/openwrt/luci/pull/2817

SvenRoederer: got merged upstream: https://github.com/openwrt/luci/pull/2817#issuecomment-660873641

PR #205, opened 2020-06-19
addons: fix building target images with openwrt imagebuilder

PolynomialDivision:
Check if we are on "real hardware" or compiling an image with imagebuilder. If we are using imagebuilder the `/lib/functions.sh` is not existing.

This should fix https://github.com/freifunk-berlin/firmware/issues/810.

### DIFF
--- a/addons/collectd-addons/Makefile
+++ b/addons/collectd-addons/Makefile
@@ -83,7 +83,9 @@ endef
 
 define Package/collectd-dnsmasq-addon/postinst
 #!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] || {
 $${IPKG_INSTROOT}/tmp/collectd-dnsmasq-addon_postinst.sh
+}
 endef
 
 $(eval $(call Download,ubnt_mibs))
--- a/addons/freifunk-berlin-bbbdigger/Makefile
+++ b/addons/freifunk-berlin-bbbdigger/Makefile
@@ -48,7 +48,9 @@ endef
 
 define Package/freifunk-berlin-bbbdigger/postinst
 #!/bin/sh
+[ -n "$$IPKG_INSTROOT" ] || {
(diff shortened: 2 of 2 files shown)


PolynomialDivision:
> beside my comments on $IPKG_INSTROOT doing it your way, will stop the code to be executed when building with imagebuilder. But when will this script be executed on the real hardware? I think with this change it will prevent execution at all in the imagebuilder and on real hardware. Which will not enable the plugin in the final consequence.
> 
> EDIT: https://github.com/freifunk/openwrt-packages/pull/22/files shows a similar usecase.

When are this post-install scripts executed when using the freifunk firmware image builder? I thought this scripts are just for opkg?

Further you mean we should do something like:
```
define Package/freifunk-berlin-bbbdigger/postinst
[ -n "$$IPKG_INSTROOT" ] || {
#!/bin/sh
$${IPKG_INSTROOT}/tmp/freifunk-berlin-bbbdigger_postinst.sh
}
endef
```

SvenRoederer:
> When are this post-install scripts executed when using the freifunk firmware image builder? I thought this scripts are just for opkg?

As running the imagebuilder fails when running this postinst directive, it's obvious that it runs these scripts. When doing a runtime-installation via opkg, the same scripts are triggered in its postinst-hook.

@Akira25 Removing the postinst, as by this test-conditon of this PR, it will fix the imagebuilder run, but I don't see a way this postinst-script will be called at all. So the setup / configuration will not happen and must be done manually be the user.

PolynomialDivision:
I made this PR without knowing that the package is only used to be installed via opkg. I'm not sure what the best solution is, but it should not break the installation via opkg? Or is this wrong syntax?

SvenRoederer:
But the package will not be correctly configured when using it in the imagebuilder (currently it fails "verbose", after the change it fails "silently"). As it stays broken either way, I prefer "failing verbose" than having the user to dig around why the setup is failing.

PR #206, opened 2020-06-23
luci-app-ffwizard-berlin: fix insufficient permissions

PolynomialDivision:
The app tries to access the freifunk community files that are placed at /etc/config/profile_X. If the lua script tries to collect the names it will get an "permission denied" error. The PR gives the lua script the permissions to access the config files.

Fixes: https://github.com/freifunk-berlin/firmware/issues/816

### DIFF
--- a/utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
+++ b/utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
@@ -1,0 +1,80 @@
+{
+        "luci-app-ffwizard-berlin": {
+                "description": "Provides access to config files and scripts",
+                "read": {
+                        "cgi-io": [
+				"exec"
+			],
+                        "uci": [
+                                "profile_*",
+                                "system",
+                                "ffwizard",
+                                "dhcp",
+                                "olsrd",
+                                "olsrd6",
+                                "firewall",
+                                "freifunk",
+                                "wireless",
+                                "network",
+                                "qos",
+                                "luci_statistics",
+                                "openvpn"
+
+                        ],
+                        "files" : {
+                                "/etc/config/ffwizard": [
+                                        "read"
+                                ],
+                                "/etc/luci-uploads/cbid.ffuplink.1.cert": [
+                                        "read"
+                                ],
+                                "/etc/openvpn/ffuplink.crt": [
+                                        "read"
+                                ],
+                                "/etc/luci-uploads/cbid.ffuplink.1.key": [
+                                        "read"
+                                ],
+                                "/etc/openvpn/ffuplink.key": [
+                                        "read"
+                                ],
+                                "/usr/libexec/rpcd/ffwizard-berlin": [
+                                        "exec"
+                                ]
+                        }
+                },
+                "write": {
+                        "uci": [
+                                "profile_*",
+                                "system",
+                                "ffwizard",
+                                "dhcp",
+                                "olsrd",
+                                "olsrd6",
+                                "firewall",
+                                "freifunk",
+                                "wireless",
+                                "network",
+                                "qos",
+                                "luci_statistics",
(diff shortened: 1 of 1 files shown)


SvenRoederer:
A similar issue is handled by https://github.com/freifunk/openwrt-packages/pull/28, where the code switches from "luci.model.uci" to libuci-lua. This lib seems to use uci directly, rather than going trough the RPCd. This seem to work for reading uci-files, but very likely not for writing to them.

As the wizard will update at least the olsr, network and wireless config, I wonder why your permissions don't list anything. Does this means all permissions or not permissions (what I would expect)?

PolynomialDivision:
> As the wizard will update at least the olsr, network and wireless config, I wonder why your permissions don't list anything. Does this means all permissions or not permissions (what I would expect)?

I think I added all relevant permissions in the new commit. :)

utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
```diff
@@ -0,0 +1,80 @@
...
+                                "firewall",
+                                "freifunk",
+                                "wireless",
+                                "network",
+                                "qos",
+                                "luci_statistics",
+                                "openvpn"
+
+                        ],
+                        "files" : {
```
SvenRoederer, line 24: Do we really need all this permissions? 
Why having "/etc/config/ffwizard" listed explicitly

utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
```diff
@@ -0,0 +1,80 @@
...
+                                        "read"
+                                ],
+                                "/usr/libexec/rpcd/ffwizard-berlin": [
+                                        "exec"
+                                ]
+                        }
+                },
+                "write": {
+                        "uci": [
+                                "profile_*",
```
SvenRoederer, line 47: AS far as I know, we only read the profiles_ only

utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
```diff
@@ -0,0 +1,80 @@
...
+                                "freifunk",
+                                "wireless",
+                                "network",
+                                "qos",
+                                "luci_statistics",
+                                "openvpn"
+                        ],
+                        "files" : {
+                                "/etc/config/ffwizard": [
+                                        "write"
```
SvenRoederer, line 63: Will giving access via UCI-key no be enough? Why explicitly this file?

utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
PolynomialDivision: Because we access this file directly and not via uci. I don't know why... Just looked into the code which files are accessed.

utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
PolynomialDivision: You are right. Thanks! I will change that.

utils/luci-app-ffwizard-berlin/root/usr/share/rpcd/acl.d/luci-app-ffwizard-berlin.json
PolynomialDivision: Maybe you are right. I was scared that I won't grant enough permissions. Maybe I should remove this again.

PR #208, opened 2020-06-25
luci-mod-freifunk-ui: fix missing dependency

PolynomialDivision:
If you install the luci-mod-freifunk-ui and click on 'Administration' you will face the erro 'module 'luci.cbi' not found'.

Fix the error by adding luci-compat as dependency.

Fixes: https://github.com/freifunk-berlin/firmware/issues/820


SvenRoederer:

PR #209, opened 2020-06-25
luci-mod-freifunk-ui: add rpcd permissions

PolynomialDivision:
Add the necessary permissions for accessing the files and scripts.

A part of: https://github.com/freifunk-berlin/firmware/issues/818

**THIS IS A DRAFT, PERMISSIONS HAVE TO BE REMOVED**


SvenRoederer:

PR #210, opened 2020-06-25
luci-mod-freifunk-ui: fix collecting of community names

PolynomialDivision:
You can not give uci a full path.

Fixes: https://github.com/freifunk-berlin/firmware/issues/821

### DIFF
--- a/utils/luci-mod-freifunk-ui/Makefile
+++ b/utils/luci-mod-freifunk-ui/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 LUCI_TITLE:=Freifunk Public and Admin LuCI UI
 LUCI_DEPENDS:=+luci-mod-admin-full +luci-lib-json +community-profiles +luci-lib-ipkg
 LUCI_PKGARCH:=all
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 include ../../freifunk-berlin-generic.mk
 
--- a/utils/luci-mod-freifunk-ui/luasrc/model/cbi/freifunk/basics.lua
+++ b/utils/luci-mod-freifunk-ui/luasrc/model/cbi/freifunk/basics.lua
@@ -15,10 +15,10 @@ community.rmempty = false
 
 local profile
 for profile in fs.glob(profiles) do
-	local name = uci:get_first(profile, "community", "name") or "?"
-	community:value(string.gsub(profile, "/etc/config/profile_", ""), name)
+	local n = string.gsub(profile, "/etc/config/profile_", "")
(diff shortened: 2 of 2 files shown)


SvenRoederer:
isn't in fact a copy of https://github.com/freifunk/openwrt-packages/commit/d1e6af6493c8d9820ed2c901b6ecc2eb50e120e5#diff-bf1a3d268cb39c4cfa542ff0d09aee56?

PolynomialDivision:
> isn't in fact a copy of [freifunk/openwrt-packages@d1e6af6#diff-bf1a3d268cb39c4cfa542ff0d09aee56](https://github.com/freifunk/openwrt-packages/commit/d1e6af6493c8d9820ed2c901b6ecc2eb50e120e5#diff-bf1a3d268cb39c4cfa542ff0d09aee56)?

No, since it was completely broken. So I did all the work in vain? We have 2 packages that are almost equal and the other one is in another repro? And u fixed it already in the other repro but not in this repro?

SvenRoederer:
I personally don't use our fork of it and started some work in adding our change back to "upstream" in a configurable way. So I lost track of porting over.
But seeing this code flipped a switch in the brain ...

utils/luci-mod-freifunk-ui/luasrc/model/cbi/freifunk/basics.lua
```diff
@@ -15,11 +15,11 @@ community.rmempty = false
...
 local profile
 for profile in fs.glob(profiles) do
-	local name = uci:get_first(profile, "community", "name") or "?"
-	community:value(string.gsub(profile, "/etc/config/profile_", ""), name)
+	local n = string.gsub(profile, "/etc/config/profile_", "")
+	local name = uci:get_first("profile_"..n, "community", "name") or "?"
+	community:value(n, name)
 end
 
-
```
SvenRoederer, line 22: As the whole change of the code seems exactly the same as upstream, I prefer not to remove this empty line, as long as it's also in the upstream code, to keep both versions easily comparable.

utils/luci-mod-freifunk-ui/Makefile
```diff
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 LUCI_TITLE:=Freifunk Public and Admin LuCI UI
 LUCI_DEPENDS:=+luci-mod-admin-full +luci-lib-json +community-profiles +luci-lib-ipkg
 LUCI_PKGARCH:=all
-PKG_RELEASE:=3
```
SvenRoederer, line 13: No need to bump this, as it's only a small code-change (which will create a new version-number from the git-commit-header by the Makefile-macros) and no change in the package-structure or functionality happens.

SvenRoederer:

PR #213, opened 2020-10-02
dnsmasq ignore br-wan ffuplink_wan

everloop2:
**What is the problem?**
syslog is spammed by: daemon.warn dnsmasq-dhcp: DHCP packet received on xxx_wan which has no address

**What is the expected behaviour?**
dnsmasq should not care / be silent for 'br-wan ffuplink_wan'

fixes freifunk-berlin/firmware#837

### DIFF
--- a/defaults/freifunk-berlin-dhcp-defaults/uci-defaults/freifunk-berlin-dhcp-defaults
+++ b/defaults/freifunk-berlin-dhcp-defaults/uci-defaults/freifunk-berlin-dhcp-defaults
@@ -5,8 +5,13 @@ guard "dhcp"
 
 # quieten down dnsmasq a bit (do not log lease-mgmt)
 uci set dhcp.@dnsmasq[0].quietdhcp=1
+
 # enable DHCP-rapidcommit (RFC4039)
 uci set dhcp.@dnsmasq[0].rapidcommit=1
+
+# dnsmasq should not care for 'br-wan ffuplink_wan', prevents: daemon.warn dnsmasq-dhcp: DHCP packet received on wan which has no address
+uci add_list dhcp.@dnsmasq[0].notinterface='br-wan'
+uci add_list dhcp.@dnsmasq[0].notinterface='ffuplink_wan'
 
 # on IPv6-islands we also should give a default-route to the clients,
 # so they can also reach IPv6-neighbours. 


SvenRoederer:
thanks for this patch. I just saw that I tried a similar thing on 2 of my nodes. 
In contrast to you I used the "nonwildcard" option of dnsmasq. Not binding to "0.0.0.0" seems to fix this also.

So not sure if it's better to 
* exclude some interface --> ```--except-interface``` of dnsmasq
* include relevant interfaces only (done by the dnsmasq init-script) --> ```--bind-dynamic``` of dnsmasq

from the dnsmasq-manual it's hard to say, which way might have more drawbacks.

> -i, --interface=<interface name>
    Listen only on the specified interface(s). Dnsmasq automatically adds the loopback (local) interface to the list of interfaces to use when the --interface option is used. If no --interface or --listen-address options are given dnsmasq listens on all available interfaces except any given in --except-interface options. On Linux, when --bind-interfaces or --bind-dynamic are in effect, IP alias interface labels (eg "eth1:0") are checked, rather than interface names. In the degenerate case when an interface has one address, this amounts to the same thing but when an interface has multiple addresses it allows control over which of those addresses are accepted. The same effect is achievable in default mode by using --listen-address. A simple wildcard, consisting of a trailing '*', can be used in --interface and --except-interface options. 
-I, --except-interface=<interface name>
    Do not listen on the specified interface. Note that the order of --listen-address --interface and --except-interface options does not matter and that --except-interface options always override the others. The comments about interface labels for --listen-address apply here. 
-z, --bind-interfaces
    On systems which support it, dnsmasq binds the wildcard address, even when it is listening on only some interfaces. It then discards requests that it shouldn't reply to. This has the advantage of working even when interfaces come and go and change address. This option forces dnsmasq to really bind only the interfaces it is listening on. About the only time when this is useful is when running another nameserver (or another instance of dnsmasq) on the same machine. Setting this option also enables multiple instances of dnsmasq which provide DHCP service to run in the same machine. 
--bind-dynamic
    Enable a network mode which is a hybrid between --bind-interfaces and the default. Dnsmasq binds the address of individual interfaces, allowing multiple dnsmasq instances, but if new interfaces or addresses appear, it automatically listens on those (subject to any access-control configuration). This makes dynamically created interfaces work in the same way as the default. Implementing this option requires non-standard networking APIs and it is only available under Linux. On other platforms it falls-back to --bind-interfaces mode. 

From the manual above the ```--interface=```  and the ```--bind-interfaces``` seem at least no good choice for our systems, because they will reduce the "automagic".

---
Using the here suggested exclude-way, seems at least the less intrusive to the current system.

SvenRoederer:
Just had a review, but the ```nonwildcard``` did not solve the issue at all.

Even adding "ffuplink_wan" to the list is not relevant for the tunnelberlin-* uplinks, we can do, as it should not have side-effects on these types.

SvenRoederer (approved):
Even adding "ffuplink_wan" to the list is not relevant for the tunnelberlin-* uplinks, we can do, as it should not have side-effects on these types.

PR #214, opened 2020-10-23
addons/snmp: correct some labels for Ubiquiti AirMax

SvenRoederer:
Running SNMP-Monitoring of a NanoBeam AC (http://monitor.berlin.freifunk.net/host.php?h=Verklaerung-core&p=snmp) shows that some labels are wrong / misleading, so fix this.

* change rx-power to rx-level
* suffix tx-power by unit "dBm"

update airMax-labels

### DIFF
--- a/addons/collectd-addons/files/collectd-snmp/ubnt-airmax.conf
+++ b/addons/collectd-addons/files/collectd-snmp/ubnt-airmax.conf
@@ -11,7 +11,7 @@ LoadPlugin snmp
     <Data "rf_signalstrength">
 	Type "signal_power"
 	Table false
-	Instance "rx-power"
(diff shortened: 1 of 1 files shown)


PolynomialDivision:
Could u post your config? I don't get the stuff running. :/
Could u post an output of snmpwalk? I really don't understand why I have those problems.

SvenRoederer:
Does the initial commit message help (https://github.com/freifunk-berlin/firmware-packages/commit/c78d9d9ffa3697998746e776374c19feae90b9e0)?
Then only modify "<Host "nodename in Freifunk-Monitor">" and "Address "<IP-address of node>""

everloop2 (approved):

PR #216, opened 2020-12-06
[WIP] owm: drop runtime dependency to rpcd

SvenRoederer:
The owm commandline tool and the common library depending on a running rpcd.
This dependency is caused by using the luci.model.uci package. As the tool runs from the commandline anyway, switch to the uci package which provides the same functionality with nearly the same API but without needing the rpcd.

I found this by working on a stripped down image for 8/32 MB devices.


SvenRoederer:
merged in https://github.com/freifunk/openwrt-packages/commit/b6270f83699fc44bf3977d61238ae4fd7f9528c4

PR #188, opened 2019-05-25
reuse LUCI build-support

SvenRoederer:
This PR implements the use of "luci.mk" from the LUCI feed to reuse its macros and definitions. So the packages-makefiles can be kept a bit more simple and use features like SRCDIET and automatic build of translations.
Some of our packages will be changed to use versionnumbering based on the githash, which removes the need for taking care of this "detail".

This also fixes https://github.com/freifunk-berlin/firmware/issues/584

PR #219, opened 2021-05-02
adapt to extended features to customize build via luci.mk

SvenRoederer:
in order shorten and simplify the package-Makefiles and esp. to avoid the manual pacakge-versioning a small wrapper to LuCIs luci.mk was addedn in https://github.com/freifunk-berlin/firmware-packages/pull/188 (c21296623dc0892245f415aaa039d0413659c339). For correct build https://github.com/openwrt/luci/pull/2637 was made, which had recently to be rewritten to get accepted. ~~Even it's not yet merged to Luci, let's adapt to this LuCI-PR, to make our packages build correctly.~~
~~A separate PR in the firmware has been made to provide the LuCI-Pr functionality as long as it's not merged upstream.~~


SvenRoederer:
The PR has been merged to Openwrt master and OpenWrt-21.02 recently.

PR #220, opened 2021-05-23
wizard: fix showing empty tabs on top of app

SvenRoederer:
I found this fix some time ago and forgot about it, as I use the wizard only rarely. 
Use nil instead of "", as caption for the individual pages of the wizard. The interpretation by LuCI seems to have changed by the rework in 19.07 to the JS-based model.

fixes https://github.com/freifunk-berlin/firmware/issues/801


SvenRoederer:
As mentioned I've tested it for some long time. I'll merge, before it starts to conflict ...

tested with themes "Bootstrap" and "Freifunk-generic" on OpenWrt-21.02

